PR #428, opened 2018-11-15
Signing a bundle now imply resolving image references into immutable references

simonferquel:
This aims at having signed bundles truly immutable. That means that all referenced images (invocationImages or images) without a digest are modified into digested references:
- if an image is available in the docker engine, and it has already a RepoDigest, we use it
- if an image is not available in the docker engine, we pull it and get its RepoDigest
- if an image is available in the docker engine, but has no RepoDigest, we consider that it has been built locally and has not yet been pushed. In that case we fail, and tell the user to push it.

Signed-off-by: Simon Ferquel <simon.ferquel@docker.com>


bacongobbler:
closing in favour of #563.

PR #506, opened 2018-11-23
Dockercon duffle version

simonferquel:
This is a PR for keeping track of the duffle version we'll use for the docker/app + duffle interactions @ DockerCon


simonferquel:
Closing this one, as we are converging with duffle master on the docker-app side

PR #509, opened 2018-11-23
[WIP] Add `duffle bundle docs`

radu-matei:
This PR introduces the the command `duffle bundle docs/info`, and the concept of self-contained docs for invocation images.

Currently, the spec states that _the `/cnab` directory of the invocation image MAY contain [...] a README.md or README.txt_ - if we decide to go ahead with self -contained docs for invocation images, we would modify it to be the spec-compliant equivalent of (at least) _highly recommended_.

Notes:

- this PR is currently marked as WIP because it does not yet handle all possible cases for doc files, and because of a few other items marked as `TODO` (see below).

- essentially, when executing the command, a container with the bundle's first `docker`-type invocation image is created (not run), and the documentation file is copied from the container to a temp file on disk. That file is read, then rendered as markdown on the console.

To test:
 
```console
$ duffle inspect pulumibase:0.1.0 --insecure
{
    "name": "pulumibase",
    "version": "0.1.0",
    "description": "",
    "invocationImages": [
        {
            "imageType": "docker",
            "image": "radumatei/pulumibase:0.1.0"
        }
    ],
    "images": [],
    "parameters": null,
    "credentials": null
}
$ duffle bundle docs pulumibase:0.1.0 --insecure
<rendeded markdown of https://github.com/deislabs/bundles/blob/master/pulumibase/README.md>
```

TODO:

- [ ] handle all possible documentation files
- [ ] decide if `docs` should be a directory reserved by the spec, and this command would copy it, if present.
- [ ] discuss how the visual installer would interact with this command (and the addition of flags to handle the output for renderers other than console)
- [ ] if the [PR to `samfoo/mdcat`](https://github.com/samfoo/mdcat/pull/2) is merged, switch the dependency to that - otherwise, keep the current fork.
- [ ] solve bug in renderer for `#` headers
- [ ] solve bug where the tar header is present in doc file (see below)

```
README.md
0100777 0000000 0000000 00000001356 13375572324 010553  0
ustar 00
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -119,7 +119,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:f4fff0c3c247502a13bde71b41f5b5ff373e4ba2aab32802e334d89382b57d4b"
+  digest = "1:d79af33e877079dedbea448d930d659b077011f761a255f10c2fddd4859a2c87"
   name = "github.com/docker/docker"
   packages = [
     "api",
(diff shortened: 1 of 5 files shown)


radu-matei:
Moved discussion in https://github.com/deislabs/cnab-spec/issues/118.

PR #512, opened 2018-11-26
WIP feat(*): add functional test

vdice:
PR adding functional tests similar to approach added to `deislabs/bundles` in https://github.com/deislabs/bundles/pull/40.

Not happy with duplication of test logic between the two repositories, but wondered if the benefit of having _some_ functional tests in this repo for the meantime outweighs this.  Then, we can discuss a more ideal path forward for sharing `duffle` functional tests between this and the `bundles` repo.

Thoughts?

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -1,0 +1,72 @@
+# Developer's Guide
+
+## Duffle Development
+
+### Build
+
+```shell
+make build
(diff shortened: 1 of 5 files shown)


vdice:
This branch is now quite out of date and as this area deserves a more holistic design, going to close this PR.

PR #553, opened 2018-11-30
Fix short-lived invocation-image exec

simonferquel:
This reorders docker_driver operations, aligning with what does the docker CLI (attach, and setup wait channels before starting the container)

This fix an issue with very short lived image executions, where the container was removed before attaching/waiting for execution end.


bacongobbler:
if you wouldn't mind rebasing this, we can then go ahead and merge. Thanks!

PR #566, opened 2018-12-03
docs(signing): mention digest calculation as a TODO item

### LINKED ISSUE #565: Update duffle documentation to highlight Docker image digests issues
Currently, the use of Docker to build things implies that images need to be pushed in order to calculate image digests. We are also currently signing bundles during build.  

PR #428 updates both sign and build to push in order to calculate digests, however this is an undesirable user experience,  as it requires the user to have push access to a repository and they can't copy/paste or directly use examples.  A better solution seems to be using buildkit and containerd (#563) but this won't happen before KubeCon. 

This results in a few gaps between `duffle` and the `CNAB specification`: 

* Bundles are not populated with digests on build
* Install will not validate digests
* Import/Export will not validate digests 

Therefore, we should update the documentation/walkthroughs to mention these current limitations to duffle and mention that they will be forth coming. The user cares about these outcomes because `duffle` currently doesn't produce a bundle that is CNAB compliant (lack of digest) and missing behavior (validation on install or import/export).

bacongobbler:
closes #565

Signed-off-by: Matthew Fisher <matt.fisher@microsoft.com>

### DIFF
--- a/docs/guides/signing-guide.md
+++ b/docs/guides/signing-guide.md
@@ -46,9 +46,15 @@ The `duffle build` command will automatically sign your bundle with the first ke
 $ duffle build -u test1@example.com
 ```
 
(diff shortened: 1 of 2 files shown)


squillace:
@vdice I leave it to you to merge. I'm only the emergency merger. :-)

Issue #527, opened 2018-11-28
validate input for `duffle build`


bacongobbler:
As we have changed the fields on duffle.toml mulitple times during the PoC, we have a lot of TOML files that are invalid input to `duffle build`, but `duffle build` will still silently accept that input without validation, ending up in weird scenarios like this one:

duffle.toml:

```toml
name = "wordpress"
registry = "cnab"
components = ["cnab"]
version = "0.2.0"
```

```console
$ duffle build .
==> Successfully built bundle wordpress:
```

Instead, `duffle build` should fail when it doesn't find fields that are required from duffle.toml.

radu-matei:
This is indeed a duplicate of #363. 

Closing.

PR #583, opened 2018-12-06
fix(index.go): warn on malformed repositories.json

bacongobbler:
due to #527, @itowlson ran into a case where a bundle with no version was added to repositories.json, causing a panic on line 152. His only recourse was to manually remove that field from repositories.json.

Instead of panicking, this PR demotes it to a warning and carries on parsing the tree.

Signed-off-by: Matthew Fisher <matt.fisher@microsoft.com>

### DIFF
--- a/pkg/repo/index.go
+++ b/pkg/repo/index.go
@@ -4,13 +4,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"sort"
 
 	"github.com/Masterminds/semver"
+	log "github.com/sirupsen/logrus"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


itowlson:
I have mixed feelings about this.  I'm glad that we now carry on past the malformed version, but I have two concerns:

1. It's noisy.  When I try to use `foo:0.1.0` it insists on telling me something that's bugging it about `foo:snodgrass`.  Shut up, computer!  I don't care about snodgrass right now!
2. It's not actionable.  Whoa, I was trying to install `foo:0.1.0` and I got this scary warning about a malformed snodgrass.  So... what do I do?  Is there a way to fix it?  How do I make it shut up?  Even if the answer is 'raise a GitHub issue and we'll tell you how to ~edit your registry~ edit the JSON file'...

itowlson:
I like demoting it to verbose.  That opts me in to hearing about all the interesting thing the program has found on its journey through the code, even those that may not be relevant to my task at hand, which is a perfect fit.

PR #602, opened 2018-12-07
docs(examples/multi-component): rm deprecated multi-component example

vdice:
Perhaps we're due for a more interesting muilti-component example bundle... in the meantime, this gets the current one dusted off/updated.


radu-matei:
I would honestly be ok with deleting this sample for now, as the builder is in semi-stable state - it should be able to build _only_ invocation images, while here the example is showing the building of application images. This is something that needs to be updated in the builder, and I think this example only confuses a new reviewer of this repo.

It should be the job of other tools to build application images, not that of `duffle`.

Thoughts?

technosophos:
I think I agree with @radu-matei 

Very glad you caught this, though. It is definitely on the dusty side.

vdice:
Roger; yup, no qualms in just deleting; will update 👍

PR #605, opened 2018-12-11
Moving to image map injection

simonferquel:
This PR changes the way we handle service images, both in term of the representation in bundle.json format, and how we deal with this information at runtime. This is also related to how we will use OCI ImageIndex to store bundles in a meaningfullway in an OCI registry, providing support for retagging, image promotion flows, mirroring, consistency of pull-rights across all services of an application etc.

- it changes the `images` field of a bundle from a flat list to a map of logical-name -> image description
- it removes the `refs` field in `image` struct
- it replace that by the injection of this map in a well known location (`/cnab/app/image-map.json`) in the invocation image, just like we do with parameters and credentials.

This way, a CNAB invocation image can consume this map to handle replacement of image references.

Let us take the case of docker-app packages:
- When docker-app creates a bundle, it will populate this map with something like:
```json
{
  "service1": {
    "image": "my-repo:my-tag"<…>
  }
}
```
- When pushing the app as a OCI Index in a registry, we will actually copy the manifest at `my-repo:my-tag` into the application repository, and reference it in the index manifests list (blobs themselves will just be mounted to avoid data-duplication, and enable image layers sharing).
- The application can then be retagged, promoted etc. These actions will all transitively copy all referenced images manifests, just by traversing manifests descriptors.
- When we pull the application from a reference like "my-registry/my-namespace/my-app:1.0.0", the transformation to the "bundle.json" representation will have something like:
```json
{
  "service1": {
    "image": "my-registry/my-namespace/my-app@<digest of this particular service image manifest>"
  }
}
```
- On installing, docker-app will consume this map, and will accordingly configure the service to use this digested reference. This way, any registry user having pull rights for the application will be guaranteed to have pull rights to pull each of its services.


simonferquel:
@technosophos The idea here, is to move the responsibility of doing image name replacements from the CNAB runtime to the invocation image itself (see https://github.com/docker/app/pull/434/files#diff-36c6b9d9df97d028267f309951037d97R53 too look at what docker-app uses it to modified the rendered compose-file so that it uses images from the bundle manifest). It solves the issue with heavily templated files (like some helm charts) which are not valid yaml/json/toml at rest - and thus, refs replacements wont work for them.
We could still keep both I think, if it makes better sense for simple invocation images (both injecting the map, and doing the "declarative" refs replacements).

technosophos:
Ah! I understand. Okay, let me think about that. I don't immediately see any drawbacks, and that does seem to substantially simplify what a CNAB client needs to be able to do.

simonferquel:
I think the hybrid approach (declarative substitution + image map injection) might be very good as well. Need some concensus here though :)

simonferquel:
Note: https://github.com/deislabs/cnab-spec/pull/62 is a PR on the CNAB spec for the hybrid approach (both declarative substitution and image map injection supported by the runtime)

simonferquel:
I'll make changes today to align to the spec (that claims support for both declarative substitution, and image map injection)

simonferquel:
@technosophos just rebased and aligned with the spec (restoring the declarative refs). We can always make a PR to remove the refs field if we decide to remove it from the spec later (still undecided AFAIK).

PR #612, opened 2019-01-14
feat: use Canonical JSON

technosophos:
This switches Duffle to use Canonical JSON, which provides JSON ordering
that is conducive to hashing. http://wiki.laptop.org/go/Canonical_JSON

This will increase interoperability with other languages and platforms,
and match the ordering produced by security tools like Notary.

Note that this change does not impact:

- Claims (still stored as regular JSON)
- Drivers (still receive regular JSON)
- Repo (still uses regular JSON)

### DIFF
--- a/pkg/duffle/manifest/create.go
+++ b/pkg/duffle/manifest/create.go
@@ -1,10 +1,11 @@
 package manifest
 
 import (
-	"encoding/json"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+
+	"github.com/docker/go/canonical/json"
 
 	"github.com/deislabs/duffle/pkg/bundle"
 )
@@ -62,7 +63,7 @@ func Scaffold(path string) error {
 		},
 	}
 
-	d, err := json.MarshalIndent(m, "", "    ")
+	d, err := json.MarshalCanonical(m)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


pkg/duffle/manifest/create.go
```diff
@@ -1,11 +1,12 @@
 package manifest
 
 import (
-	"encoding/json"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
+	"github.com/docker/go/canonical/json"
```
radu-matei: nit: Should the order of the two imports be switched?

pkg/signature/signature.go
```diff
@@ -3,10 +3,11 @@ package signature
 import (
 	"bytes"
 	"crypto"
-	"encoding/json"
 	"errors"
 	"io"
 
+	"github.com/docker/go/canonical/json"
```
radu-matei: nit: Should this be the last imported package?

pkg/duffle/manifest/create.go
technosophos: I don't know what the expected order is anymore. I basically reordered until the linter stopped whining.

pkg/signature/signature.go
radu-matei: I'm sure I was wrong here, so marking as resolved.

radu-matei:
Now that we moved passed the order import kerfuffle, this is tied to https://github.com/deislabs/cnab-spec/pull/73.

It also needs a rebase.

technosophos:
Rebased.

PR #613, opened 2019-01-16
Fix not installing when invocation image not present

### LINKED ISSUE #596: bug: `duffle install` doesn't pull invocation image if not present locally
Executing a `duffle install` when the bundle invocation image was not already pulled results in a error:

```
$ duffle install pull-test -f bundle.json --insecure
Executing install action...
Unable to find image 'cnab/example-credentials:latest' locally
```

radu-matei:
On Debian at least, it turns out we are hitting https://github.com/Nvveen/Gotty/pull/1 - workaround is to pass `isTerm = false` to `jsonmessage.DisplayJSONMessagesStream()`.

closes #596

### DIFF
--- a/pkg/driver/docker_driver.go
+++ b/pkg/driver/docker_driver.go
@@ -83,12 +83,9 @@ func pullImage(ctx context.Context, cli command.Cli, image string) error {
 		return err
 	}
 	defer responseBody.Close()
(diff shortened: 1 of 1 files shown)


radu-matei:
Failing linter errors should be solved by #612. I think?

radu-matei:
@michelleN - rebased.

PR #615, opened 2019-01-23
ref(*) update export to support bundle reference

michelleN:


cmd/duffle/export_test.go
```diff
@@ -0,0 +1,113 @@
+package main
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	//"github.com/deislabs/duffle/pkg/bundle"
```
vdice: Can this line be removed now?

michelleN: yes! thanks

michelleN:
@vdice Updated the docs! Good catch. Thanks.

PR #618, opened 2019-01-29
ref(*): decouple imagebuilder from bundle builder

michelleN:


pkg/builder/builder.go
```diff
@@ -142,26 +136,26 @@ func (b *Builder) version(baseVersion, sha string) (string, error) {
...
-	if err := buildComponents(ctx, app); err != nil {
-		return fmt.Errorf("error building components: %v", err)
+	if err := buildInvocationImages(ctx, b.ImageBuilders, app); err != nil {
+		return fmt.Errorf("error building image: %v", err)
 	}
 	return nil
 }
 
-func buildComponents(ctx context.Context, app *AppContext) (err error) {
+func buildInvocationImages(ctx context.Context, imageBuilders []imagebuilder.ImageBuilder, app *AppContext) (err error) {
```
radu-matei: Nice catch on renaming the method :+1:

pkg/builder/builder.go
```diff
@@ -85,32 +76,35 @@ func (b *Builder) PrepareBuild(bldr *Builder, mfst *manifest.Manifest, appDir st
...
+		}
+		newver, err := b.version(baseVersion, strings.Split(imb.URI(), ":")[1])
+		if err != nil {
+			return nil, nil, err
 		}
+		bf.Version = newver
+	}
+
+	for _, image := range mfst.Images {
+		//TODO don't need imagebuilders for images anymore so not sure what this looks like
```
radu-matei: Let me see if I can recall the context for this:

- we decided that the builder is only responsible for building invocation images
- as discussed in #564, we also want to add other images to the bundle (for example images we don't build / are already built)
- we could use this to iterate on the Duffle manifest referenced images and add them to the bundle?

PR #623, opened 2019-02-04
ref(cmd/duffle): export bundle ref or file

### LINKED ISSUE #582: Export from file system should take a file not a directory
At the moment 'duffle export foo' treats foo as a directory, and looks for a bundle.json or bundle.cnab according to whether the '--insecure' flag is passed.  For consistency with othe commands, I suggest:

* export foo: exports from local store (and supports version)
* export -f path: exports from filesystem and treats path as a file name not a directory and infers signedness from file format
* --insecure flag skips signature verification for signed bundles, does nothing on unsigned bundles

Thoughts?  cc @michelleN

michelleN:
duffle export takes one arg. It figures out if arg is
filepath or a bundle reference and then exports accordingly

resolves #582

### DIFF
--- a/cmd/duffle/export_test.go
+++ b/cmd/duffle/export_test.go
@@ -53,15 +53,15 @@ func TestExportSetup(t *testing.T) {
 
 	duffleHome := home.Home(tempDuffleHome)
 	exp := &exportCmd{
-		bundleRef: "foo:1.0.0",
-		dest:      tempDir,
-		home:      duffleHome,
-		out:       out,
+		bundle: "foo:1.0.0",
+		dest:   tempDir,
+		home:   duffleHome,
+		out:    out,
 	}
 
 	source, _, err := exp.setup()
 	if err != nil {
-		t.Fatal(err)
+		t.Errorf("Did not expect error but got %s", err)
 	}
 
 	expectedSource := filepath.Join(tempDuffleHome, "bundles", "foo-1.0.0.cnab")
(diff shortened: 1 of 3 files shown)


cmd/duffle/export_test.go
```diff
@@ -53,15 +53,15 @@ func TestExportSetup(t *testing.T) {
...
+		bundle: "foo:1.0.0",
+		dest:   tempDir,
+		home:   duffleHome,
+		out:    out,
 	}
 
 	source, _, err := exp.setup()
 	if err != nil {
-		t.Fatal(err)
+		t.Errorf("Did not get expect error but got %s", err)
```
vdice: nit: slight typo here in this message... extra `get` maybe could be removed?

cmd/duffle/export_test.go
```diff
@@ -70,16 +70,31 @@ func TestExportSetup(t *testing.T) {
...
+	bundlepath := filepath.Join("..", "..", "tests", "testdata", "bundles", "foo.json")
+	expFile := &exportCmd{
+		bundle: bundlepath,
+		dest:   tempDir,
+		home:   duffleHome,
+		out:    out,
+	}
+	source, _, err = expFile.setup()
+	if err != nil {
+		t.Errorf("Did not get expect error but got %s", err)
```
vdice: same typo in error message as mentioned above

michelleN:
thanks for the review @vdice !

Issue #577, opened 2018-12-05
`duffle creds generate` asks values for credentials in a random order


dgageot:
+ Create a `bundle.json` with two credentials: `cred1` and `cred2`.
 + Run `duffle creds generate NAME -f ./bundle.cnab`. It'll ask values for both credentials but in a random order.

PR #627, opened 2019-02-08
Sorts credentials alphabetically before prompting the user during generation

youreddy:
Resolves https://github.com/deislabs/duffle/issues/577

### DIFF
--- a/cmd/duffle/credential_generate.go
+++ b/cmd/duffle/credential_generate.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cmd/duffle/credential_generate.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/deislabs/duffle/pkg/bundle"
 	"github.com/deislabs/duffle/pkg/credentials"
 	"github.com/deislabs/duffle/pkg/duffle/home"
+	"sort"
```
radu-matei: This makes the linter unhappy:

```
golangci-lint run --config ./golangci.yml
cmd/duffle/credential_generate.go:20: File is not `goimports`-ed (goimports)
```

youreddy: Ahh yes, setting up the linter now. I'll resubmit this PR soon!

Issue #629, opened 2019-02-11
Implement `stateless` flag from the spec


technosophos:
Add support for `stateless`:

https://github.com/deislabs/cnab-spec/pull/85

I think this should be fairly straightforward, but probably not easy enough to be a good first issue.

radu-matei:
cc @simonferquel

radu-matei:
As initial support for this lands in #630, this brings an interesting discussion regarding the duffle UX:

- according to the spec, [a stateless] _action does not act on a claim_.
- custom actions are run at this point with `duffle run`: `duffle run ACTION RELEASE_NAME [flags]`.

This means that we would need `duffle run` to be able to act both on a release, and on a bundle.

What would be the best way to ensure we keep the command clear?
If we think documenting is enough, I'd also be happy with that, but we need to clearly state _how_ actions can be rung with Duffle.

astrieanna:
This and https://github.com/deislabs/duffle/issues/830 seem like duplicates.

PR #630, opened 2019-02-12
Add support for stateless actions

simonferquel:
This implements stateless actions as defined in https://github.com/deislabs/cnab-spec/pull/85.

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -20,7 +20,7 @@ func (i *Install) Run(c *claim.Claim, creds credentials.Set, w io.Writer) error 
 		return err
 	}
 
-	op, err := opFromClaim(claim.ActionInstall, c, invocImage, creds, w)
+	op, err := opFromClaim(claim.ActionInstall, false, c, invocImage, creds, w)
 	if err != nil {
 		return err
 	}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -20,7 +20,7 @@ func (i *Install) Run(c *claim.Claim, creds credentials.Set, w io.Writer) error 
 		return err
 	}
 
-	op, err := opFromClaim(claim.ActionInstall, false, c, invocImage, creds, w)
+	op, err := opFromClaim(claim.ActionInstall, notStateless, c, invocImage, creds, w)
 	if err != nil {
 		return err
 	}
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -120,7 +120,7 @@ type Action struct {
 	// Stateless indicates that the action is purely informational, that credentials are not required, and that the runtime should not keep track of its invocation
 	Stateless bool `json:"stateless,omitempty" mapstructure:"stateless"`
 	// Description describes the action as a user-readable string
-	Description string `json:"description,omitempty" mapstructure:"description"`
+	Description string `json:"description,omitempty" mapstructure:"description,omitempty"`
 }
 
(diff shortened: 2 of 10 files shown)


radu-matei:
Ok, this is in more advanced state than https://github.com/radu-matei/duffle/tree/feat-629-sync-action-spec, so I'll close that one.

Just for future reference, let's make sure we assign an issue before starting the work on it, so we don't duplicate work 😃

pkg/bundle/bundle.go
```diff
@@ -116,7 +116,13 @@ type Action struct {
 	// Modifies indicates whether this action modifies the release.
 	//
 	// If it is possible that an action modify a release, this must be set to true.
-	Modifies bool
+	Modifies bool `json:"modifies,omitempty"`
```
radu-matei: Any specific reason why this doesn't also include the `mapstructure:"modifies"` tag?

Same for `Stateless` and `Description`.

pkg/action/install.go
```diff
@@ -20,7 +20,7 @@ func (i *Install) Run(c *claim.Claim, creds credentials.Set, w io.Writer) error
 		return err
 	}
 
-	op, err := opFromClaim(claim.ActionInstall, c, invocImage, creds, w)
+	op, err := opFromClaim(claim.ActionInstall, false, c, invocImage, creds, w)
```
radu-matei: This is more of a nit, but would it increase readibility if the `false` here passed for `stateless` was a constant?

Same for the other places where `opFromClaim` is used without directly passing the action stateless field.

pkg/bundle/bundle.go
simonferquel: My bad, will fix that

pkg/action/install.go
simonferquel: good call

simonferquel:
Sorry about the duplicated work. Fixed the annotations and added a `notStateless` const for more clarity

radu-matei:
This looks good, although I have a question (that I think will result in an issue for Duffle) - according to the spec for stateless actions: _The action does not act on a claim_ - however, `duffle run` requires an installation name, and does not accept a bundle.

If I understand the usage of stateless actions correctly, the above would be something we want to support, right?

simonferquel:
Yes, I only focused on the lower level things for now (as my honest goal was to remove a hack we have in docker-app which already defines implicitly stateless actions).
I have no idea how the duffle UX should be for such action though. Docker-app is easier as it is an opinionated tool and we have top level commands like `docker app inspect/render` for them.
________________________________

simonferquel:
Hmm, also, we must also be able to pass parameter values for such stateless actions like we do with `docker app render -p nginx.scale=3`. Definitely a tricky UX problem.

radu-matei:
Yeah, that totally makes sense, and I'll create a new issue to track this.

Could you please, also sign your commits?
Otherwise, this LGTM.

simonferquel:
They are signed with `git commit -s`. Do you require a PGP signature ?

radu-matei:
Yeah, we haven't enforced this yet, but as it's best practice, I think we should start to.

radu-matei:
Oh, there's a small conflict on this - once it's resolved, I think we can go ahead and merge this, keeping the discussion about the UX issue in #629.

Thanks!

simonferquel:
Oops, sorry, totally forgot about this one. Rebasing right now

simonferquel:
Now rebased, PTAL @radu-matei

PR #631, opened 2019-02-12
Adds alias `duffle bundle show` of existing command `duffle show`

ryanmoran:
* Refactors the contents of the `duffle show` command out into a struct.
* Invokes the common struct in both `duffle show` and `duffle bundle show` commands.

Resolves https://github.com/deislabs/duffle/issues/535

### DIFF
--- a/cmd/duffle/bundle_show.go
+++ b/cmd/duffle/bundle_show.go
@@ -1,0 +1,81 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+func newBundleShowCmd(w io.Writer) *cobra.Command {
+	bsc := &bundleShowCmd{}
+	bsc.w = w
+
+	cmd := &cobra.Command{
+		Use:   "show NAME",
+		Short: "return low-level information on application bundles",
+		Long:  bsc.usage(true),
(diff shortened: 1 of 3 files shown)


cmd/duffle/bundle_show.go
```diff
@@ -0,0 +1,73 @@
...
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+func newBundleShowCmd(w io.Writer) *cobra.Command {
+	bsc := &bundleShowCmd{}
+	bsc.w = w
+
+	const usage = ` Returns information about an application bundle.
```
radu-matei: nit: I don't really like that we duplicate the usage - whenever we'll modify something, we're going to have to modify it twice.

Could you please add this as a `const` outside the function and reference it in `show.go`?

ryanmoran: Move the shared long-form usage into a method here: https://github.com/deislabs/duffle/pull/631/files#diff-67b87ff356fa854fe2a2ca92f31e2c18R41.

With this you can indicate if the usage should show examples of the form `duffle show` or `duffle bundle show`.

radu-matei: Nice!

ryanmoran:
Latest commit should be signed. Thanks!

Issue #329, opened 2018-11-01
`duffle inspect` should accept digests


bacongobbler:
See #307 for `duffle inspect`

In #307, `duffle inspect` accepts a repository name as input. It should probably accept a digest as well.

For example:

```
$ duffle build .
...
==> Successfully built bundle helloworld:a5d8b894e648ba3dbe9553ded3f133fe75f02b86
$ duffle bundles
NAME                            VERSION                                         DIGEST
hub.cnlabs.io/helloworld        1.0.0                                           556fa7c5dd67621200414571ebeb9f3dc5d9f0da
$ duffle inspect hub.cnlabs.io/helloworld:1.0.0
-----BEGIN PGP SIGNED MESSAGE-----
...
```

But this does not work:

```
$ duffle inspect 556fa7c5dd67621200414571ebeb9f3dc5d9f0da
Error: no bundle name found
```

PR #632, opened 2019-02-12
Allows bundles to be shown given a digest

ryanmoran:
Resolves https://github.com/deislabs/duffle/issues/329

/cc @youreddy

### DIFF
--- a/pkg/repo/index.go
+++ b/pkg/repo/index.go
@@ -140,21 +140,34 @@ func (i Index) Get(name, version string) (string, error) {
 // If the name is not found, this will return false.
 func (i Index) GetVersions(name string) ([]BundleVersion, bool) {
 	ret, ok := i[name]
-	rawversions := []string{}
-	for ver := range ret {
-		rawversions = append(rawversions, ver)
+	if !ok {
+		ret, ok = i.versionsWithDigest(name)
 	}
 
-	bv := []BundleVersion{}
-	for _, r := range rawversions {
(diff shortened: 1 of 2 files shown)


pkg/repo/index.go
```diff
@@ -140,20 +140,32 @@ func (i Index) Get(name, version string) (string, error) {
 // If the name is not found, this will return false.
 func (i Index) GetVersions(name string) ([]BundleVersion, bool) {
 	ret, ok := i[name]
-	rawversions := []string{}
-	for ver := range ret {
-		rawversions = append(rawversions, ver)
+	if !ok {
+	Loop:
```
radu-matei: nit: do you have strong feelings about keeping the label here?
As it is not required in `for` statements, and not widely used, it might confuse people not familiar with labels.

But this is not necessarily something keeping this PR from getting in.

ryanmoran: I'm not immediately seeing a simple way to break through the doubly-nested loop here. My understanding of reading the [Golang spec for `break` statements](https://golang.org/ref/spec#Break_statements) is that it only drops out of the inner-most loop if not specified with a loop label. That is why I have included the `Loop` label on the outer `for` loop.

radu-matei: I guess you could extract the nested loop into a separate function and directly return when you found it.

ryanmoran:
@radu-matei I've pulled that doubly-nested loop out into a function as suggested. Thanks!

Issue #528, opened 2018-11-28
`duffle install` doesn't accept "latest" as a version specifier


bacongobbler:
Duffle should enforce users to specify a tag, as currently `latest` is not considered a valid tag. Either that or loosen the constraint to include arbitrary strings.

The current error output:

```
$ duffle install w wordpress
Error: could not find wordpress:latest in /Users/admin/.duffle/repositories.json: improper constraint: latest
```

A better error message would be:

```
$ duffle install w wordpress
Error: please specify a tag in the form of 'duffle install w wordpress:0.1.0'
```

Or, if we loosened the semver constraint and keep previous behaviour:

```
$ duffle install w wordpress
Error: could not find wordpress:latest in /Users/admin/.duffle/repositories.json
```

youreddy:
@itowlson @bacongobbler taking into the considerations suggested above, how do you all feel about this error message?:

```
$ duffle install a aks-terraform:latest
Error: 'latest' is not a valid tag option
```
I created https://github.com/deislabs/duffle/pull/636 for this. Let me know if I should update the error message in it.

michelleN:
@youreddy I think it might be clearer if we the error looked like this: `Error: 'latest' is not a valid bundle version` What do you think?

youreddy:
> @youreddy I think it might be clearer if we the error looked like this: `Error: 'latest' is not a valid bundle version` What do you think?

Sounds good to me @michelleN. I went ahead and made a quick PR for this https://github.com/deislabs/duffle/pull/653

PR #636, opened 2019-02-15
Error when the 'latest' tag is used

youreddy:
Resolves issue: https://github.com/deislabs/duffle/issues/528


youreddy:
Hey @michelleN, thank you for taking the time to look into this and the explanation. I agree that it makes sense to move this logic to the bundle `Validate()` function. 

I've pushed up changes to the branch to reflect those changes. Let me know if that's more in line with the changes requested.

PR #639, opened 2019-02-20
Proposal Implementation: Replace env var and file inputs with JSON on STDIN

ryanmoran:
This is a rough implementation of a CNAB spec [proposal](https://github.com/deislabs/cnab-spec/issues/114) to change the process for passing configuration data into invocation images. Please excuse my ignorance of the golang libraries for interacting with docker. I believe that I made the minimal set of changes required to stream input to STDIN for the run tool when performing an action.

The changes here are not really meant to be merged, but more to serve as a living example of what the changes to a runtime would be as a result of the spec proposal.

### DIFF
--- a/examples/helloworld/cnab/Dockerfile
+++ b/examples/helloworld/cnab/Dockerfile
@@ -1,7 +1,7 @@
 FROM alpine:latest
 
 RUN apk update
-RUN apk add -u bash
+RUN apk add -u bash jq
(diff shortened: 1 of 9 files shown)


ryanmoran:
With the merging of https://github.com/deislabs/cnab-spec/pull/153, this no longer makes sense. Closing.

Issue #642, opened 2019-02-21
make bundle file argument UX consistent across commands


michelleN:
see #582 for context
all commands should follow export UX. 

If a command requires a bundle as an argument, the bundle argument can be overridden by specifying a flag `--bundle-is-file`

jeremyrickard:
Closed by#643

PR #643, opened 2019-02-21
ref(cmd/duffle): install takes bundle file as arg + upgrade/uninstall UX refactor

michelleN:
* makes UX consistent with export
* addresses #642


cmd/duffle/uninstall.go
```diff
@@ -3,71 +3,73 @@ package main
...
-	flags.StringArrayVarP(&credentialsFiles, "credentials", "c", []string{}, "Specify credentials to use inside the CNAB bundle. This can be a credentialset name or a path to a file.")
-	flags.StringVarP(&bundleFile, "file", "f", "", "bundle file to install")
-	flags.StringVarP(&uc.valuesFile, "parameters", "p", "", "Specify file containing parameters. Formats: toml, MORE SOON")
-	flags.StringArrayVarP(&uc.setParams, "set", "s", []string{}, "set individual parameters as NAME=VALUE pairs")
-	flags.BoolVarP(&uc.insecure, "insecure", "k", false, "Do not verify the bundle (INSECURE)")
+	flags.StringVarP(&uninstall.driver, "driver", "d", "docker", "Specify a driver name")
+	flags.StringArrayVarP(&uninstall.credentialsFiles, "credentials", "c", []string{}, "Specify credentials to use inside the CNAB bundle. This can be a credentialset name or a path to a file.")
+	flags.StringVarP(&uninstall.valuesFile, "parameters", "p", "", "Specify file containing parameters. Formats: toml, MORE SOON")
+	flags.StringVarP(&uninstall.bundle, "bundle", "b", "", "bundle to uninstall")
+	flags.StringVar(&uninstall.bundleFile, "bundle-file", "", "path to a bundle file to upgrade")
```
vdice: `s/upgrade/uninstall` in this var description

cmd/duffle/uninstall.go
```diff
@@ -3,71 +3,73 @@ package main
...
+const uninstallUsage = `Uninstalls an installation of a CNAB bundle.
 
 When using '--parameters' or '--set', the uninstall command will replace the old
 parameters with the new ones supplied (even if the new set is an empty set). If neither
 '--parameters' nor '--set' is passed, then the parameters used for 'duffle install' will
 be re-used.
 `
 
-var uninstallDriver string
+var BundleAndBundleFileError = errors.New("Both --bundle and --bundle-file flags cannot be set")
```
vdice: Selecting this line not so much to comment on the line itself but as a convenient spot to raise the following question:

For this `uninstall` action, what are some scenarios that a user would supply a bundle (via either the `--bundle` or `--bundle-file` option) when uninstalling?  Or, putting it another way, how would this provided bundle change the uninstallation action?

It seems we'd enable a user to get into some wild situations wherein they'd install a bundle, say, using the wordpress bundle.json and then subsequently uninstall this installation with, intentionally or unintentionally, a different bundle.json (either via a different _version_ of the wordpress bundle.json or an entirely different bundle.json)

cmd/duffle/upgrade.go
```diff
@@ -24,78 +25,80 @@ If no parameters are passed, the parameters from the previous release will be us
...
 	flags := cmd.Flags()
-	flags.StringVarP(&bundleFile, "file", "f", "", "Set the bundle file to use for upgrading")
-	flags.StringVarP(&upgradeDriver, "driver", "d", "docker", "Specify a driver name")
-	flags.StringArrayVarP(&credentialsFiles, "credentials", "c", []string{}, "Specify credentials to use inside the CNAB bundle. This can be a credentialset name or a path to a file.")
-	flags.StringVarP(&uc.valuesFile, "parameters", "p", "", "Specify file containing parameters. Formats: toml, MORE SOON")
-	flags.StringArrayVarP(&uc.setParams, "set", "s", []string{}, "Set individual parameters as NAME=VALUE pairs")
-	flags.BoolVarP(&uc.insecure, "insecure", "k", false, "Do not verify the bundle (INSECURE)")
-	flags.StringArrayVarP(&uc.setFiles, "set-file", "i", []string{}, "Set individual parameters from file content as NAME=SOURCE-PATH pairs")
+	flags.StringVarP(&upgrade.driver, "driver", "d", "docker", "Specify a driver name")
+	flags.StringVarP(&upgrade.bundle, "bundle", "b", "", "bundle to use for upgrading")
```
vdice: A similar question here as posed around the `uninstall` action cmd.  

Looking to understand how supplying a bundle is intended to affect an `upgrade` action?  I think I can envision the use case a bit more clearly with upgrade: a user has installed a bundle with a given version of a bundle.json and then wishes to upgrade the installation with a revised version of the bundle.json.

If I'm on the right track, I guess my question becomes: Have we yet explored the intricacies and potentially wide-ranging margin for errors when enabling the upgrade of an installation with a different version of a bundle.json (may possibly be vastly different) than was used for the initial installation?

cmd/duffle/uninstall.go
michelleN: That's a good question. This PR is not changing existing functionality. As of now, you are able to pass in a bundle with uninstall. This is simply a refactor for UX.
An end user may want to pass in a bundle to run an uninstall action bc the new bundle or bundle version may contain updated logic for the uninstall action. The default scenario for uninstall does not require a bundle argument. Rather, this is a way for the user to override the existing uninstall action.

cmd/duffle/upgrade.go
michelleN: See comment above. I think it addresses the same question. Perhaps we should open up an issue to discuss this further.

cmd/duffle/uninstall.go
vdice: I see, that definitely helps me understand.  Thanks @michelleN ~

PR #651, opened 2019-02-26
Remove Docker socket from file mounts in driver

### LINKED ISSUE #650: Docker daemon access bug
The following issue was reported in `cnab-spec`, but is actually a bug in Duffle: https://github.com/deislabs/cnab-spec/issues/108

radu-matei:
closes #650
closes https://github.com/deislabs/cnab-spec/issues/108

This PR simply removes the Docker socket.
We had an initial discussion during last week's standup, but I'll re-start it here: with the current implementation, the bundles that deployed Docker Compose files to the same daemon no longer work in their current state.

Should there be a way to run the invocation image as a privileged container in Duffle?

### DIFF
--- a/pkg/driver/docker_driver.go
+++ b/pkg/driver/docker_driver.go
@@ -14,7 +14,6 @@ import (
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
-	"github.com/docker/docker/api/types/mount"
 	"github.com/docker/docker/api/types/strslice"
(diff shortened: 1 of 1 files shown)


simonferquel:
For Docker-app, when we detect we are deploying to the local docker-desktop engine, we'd like to be able to hook the ContainerConfig / HostConfig generated by the driver. I'll do a PR about that based on this current one (and duffle might also want to support a `--mount-docker-socket` flag as well for the Docker Compose sample)

radu-matei:
Closing as #661 included these changes.

Issue #634, opened 2019-02-13
Docs: Add instructions to download hosted binaries


vdice:
The current installation steps in duffle's [README](https://github.com/deislabs/duffle/blob/master/README.md) are around building from source.

We should add a guide on how/where to fetch pre-built/hosted binaries.  Currently, these are hosted with a corresponding release, e.g. https://github.com/deislabs/duffle/releases/tag/0.1.0-ralpha.4%2Bdramallamabuie

Issue #624, opened 2019-02-05
Explain how to get duffle and a copy of the bundles repo to match


carolynvs:
Document in a single place how to duffle installed along side the bundle examples and have them both be the right version. Both on master will not work.

PR #652, opened 2019-02-26
Update readme with release and few notes

radu-matei:
Addresses #624

As we don't have yet a clear strategy for syncing / pinning releases for Duffle and bundles, I left that part of the readme out for now.


radu-matei:
closes #634

PR #653, opened 2019-02-26
Update error message for 'latest' tag

youreddy:
Resolves https://github.com/deislabs/duffle/issues/528#issuecomment-467072831


michelleN:
Thanks @youreddy

PR #660, opened 2019-03-06
fix: Switch from ARM Template params to JSON Schema params

### LINKED ISSUE #648: Change parameters to use the JSON Schema format instead of the ARM format
See https://github.com/deislabs/cnab-spec/pull/77 for a full discussion, but in a nutshell, the fields used in parameters have changed to match the more powerful JSON Schema spec (instead of the ARM Template format).

technosophos:
Switch to JSON Schema's keywords for validators, add pattern validator and exclusive[Min|Max] validators.

Also, refactored the tests to (a) increase coverage, and (b) reduce boilerplate.

Closes #648

### DIFF
--- a/pkg/bundle/parameters.go
+++ b/pkg/bundle/parameters.go
@@ -3,23 +3,27 @@ package bundle
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 )
 
 // ParameterDefinition defines a single parameter for a CNAB bundle
 type ParameterDefinition struct {
-	DataType      string             `json:"type" mapstructure:"type"`
-	DefaultValue  interface{}        `json:"defaultValue,omitempty" mapstructure:"defaultValue"`
-	AllowedValues []interface{}      `json:"allowedValues,omitempty" mapstructure:"allowedValues"`
-	Required      bool               `json:"required,omitempty" mapstructure:"required"`
-	MinValue      *int               `json:"minValue,omitempty" mapstructure:"minValue"`
-	MaxValue      *int               `json:"maxValue,omitempty" mapstructure:"maxValue"`
-	MinLength     *int               `json:"minLength,omitempty" mapstructure:"minLength"`
-	MaxLength     *int               `json:"maxLength,omitempty" mapstructure:"maxLength"`
-	Metadata      *ParameterMetadata `json:"metadata,omitempty" mapstructure:"metadata"`
-	Destination   *Location          `json:"destination,omitemtpty" mapstructure:"destination"`
-	ApplyTo       []string           `json:"apply-to,omitempty" mapstructure:"apply-to,omitempty"`
+	DataType         string            `json:"type" mapstructure:"type"`
+	DefaultValue     interface{}       `json:"defaultValue,omitempty" mapstructure:"defaultValue,omitempty"`
+	Required         bool              `json:"required" mapstructure:"required"`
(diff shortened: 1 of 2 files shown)


pkg/bundle/parameters.go
```diff
@@ -89,6 +93,7 @@ func (pd ParameterDefinition) CoerceValue(value interface{}) interface{} {
 		if ok {
 			i, ok := asInt(f)
 			if !ok {
+				// Why do we return a float here?
```
radu-matei: Reading the `asInt` method, if the two values are not equal, the returned `i` here is 0.
I'd ask @itowlson about the logic, and if returning a float doesn't break  the validation later on.

itowlson: I believe the goal here is to coerce the read value to an integer.  If the value of `f` is integral, then `asInt` returns the corresponding integer (and true).  If `asInt` returns false, then `f` _isn't_ integral, so we can't coerce it to an integer type, but we have to return something so we return `f`.  The comment implies this should never happen because the value should already have been validated as being integral - this is solely a runtime type coercion, not a validation/conversion.

itowlson: Coming back to it at this remove, it is certainly not very self-documenting and could probably be improved with the use of error values or whatever.  And moar comments.

technosophos: LOL... I forgot that I left the comment in there. I'll rewrite this comment and merge something less weird.

technosophos:
Updated with a pointer to the explanation here.

itowlson:
Note that this is going to be a breaking change for downstream tooling, which parses the bundle's `parameters` section in order to interactively validate input values.

technosophos:
Oh! Good catch @itowlson !

radu-matei:
As this also touches on `ApplyTo` for parameters, can you please take a look, @silvin-lubecki?

Thanks!

pkg/bundle/parameters.go
```diff
@@ -3,23 +3,27 @@ package bundle
...
+	Required         bool              `json:"required" mapstructure:"required"`
+	Minimum          *int              `json:"minimum,omitempty" mapstructure:"minimum,omitempty"`
+	Maximum          *int              `json:"maximum,omitempty" mapstructure:"maximum,omitempty"`
+	ExclusiveMinimum *int              `json:"exclusiveMinimum,omitempty" mapstructure:"exclusiveMinimum,omitempty"`
+	ExclusiveMaximum *int              `json:"exclusiveMaximum,omitempty" mapstructure:"exclusiveMaximum,omitempty"`
+	MinLength        *int              `json:"minLength,omitempty" mapstructure:"minLength,omitempty"`
+	MaxLength        *int              `json:"maxLength,omitempty" mapstructure:"maxLength,omitempty"`
+	Pattern          string            `json:"pattern,omitempty" mapstructure:"pattern,omitempty"`
+	Enum             []interface{}     `json:"enum,omitempty" mapstructure:"enum,omitempty"`
+	Metadata         ParameterMetadata `json:"metadata,omitempty" mapstructure:"metadata,omitempty"`
```
michelleN: Should `Metadata` be `Description`?

radu-matei: Looking at the [parameter definition from the spec](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#parameters):

- `metadata`: Holds fields that are not used in validation (OPTIONAL)
  - `description`: A user-friendly description of the parameter

And it's the same here.

technosophos:
This is outdated.

PR #664, opened 2019-03-14
Add "apply-to" field to parameter definition

silvin-lubecki:
Also add a check that an action is in the parameter's apply-to list.

Specifications here: https://github.com/deislabs/cnab-spec/pull/103/files

### DIFF
--- a/pkg/action/action_test.go
+++ b/pkg/action/action_test.go
@@ -144,6 +144,69 @@ func TestOpFromClaim_UndefinedParams(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestOpFromClaim_MissingRequiredParameter(t *testing.T) {
+	now := time.Now()
+	b := mockBundle()
+	b.Parameters["param_one"] = bundle.ParameterDefinition{Required: true}
+
+	c := &claim.Claim{
+		Created:  now,
+		Modified: now,
+		Name:     "name",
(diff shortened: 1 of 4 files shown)


pkg/action/action_test.go
```diff
@@ -144,6 +144,32 @@ func TestOpFromClaim_UndefinedParams(t *testing.T) {
...
+			"param_one":   "oneval",
+			"param_two":   "twoval",
+			"param_three": "threeval",
+			"param_test":  "only for test action",
+		},
+	}
+	invocImage := c.Bundle.InvocationImages[0]
+
+	_, err := opFromClaim(claim.ActionInstall, c, invocImage, mockSet, os.Stdout)
+	assert.EqualError(t, err, `parameter "param_test" does not apply to action "install"`)
```
simonferquel: I think, the param should just be ignored:
if you have an action with "modify:true"  which has an action-specific parameter, this parameter value will be recorded in the claim after execution. If you then try to execute uninstall, you'll get an error `parameter "foo" does not apply to action "uninstall"`

silvin-lubecki: Ok right, but said like that it means that the `apply-to` is useless... except for `REQUIRED` parameters specific to an action. I'll update the check with that 👍

silvin-lubecki:
@simonferquel PTAL

PR #666, opened 2019-03-15
[feat-633-bundle-actions-cmd] Add bundle actions command

### LINKED ISSUE #633: Add `duffle actions` command to list actions in a bundle
```
$ duffle actions BUNDLE
```

This should list out all of the actions, and whether they are stateless, and whether they are read-only.

radu-matei:
close #633 

This PR adds a `duffle bundle actions` command that reads all actions from a bundle and displays them.

To test this, the following bundle file can be used:

```json
{
  "actions": {
    "dry-run": {
      "description": "prints what install would do with the given parameters values",
      "modifies": false,
      "stateless": true
    },
    "migrate": {
      "description": "migration action",
      "modifies": false
    },
    "status": {
      "description": "retrieves the status of an installation",
      "modifies": false
    }
  },
  "credentials": null,
  "description": "A short description of your bundle",
  "images": null,
  "invocationImages": [
    {
      "image": "padme/awesome-bundle-cnab:acff9ea424b7735b7cd6c342b7a6073f35337658",
      "imageType": "docker",
      "platform": {}
    }
  ],
  "keywords": [
    "cnab"
  ],
  "maintainers": [
    {
      "email": "padme@admidala.dev",
      "name": "Padme Amidala",
      "url": ""
    }
  ],
  "name": "awesome-bundle",
  "parameters": null,
  "version": "0.1.0"
}
```
Alternatively, the bundle can be build, and/or signed using `duffle build`.

```
$ duffle bundle actions -f bundle.json --insecure
ACTION 	STATELESS	MODIFIES	DESCRIPTION                                                  
migrate	false    	false   	migration action                                             
status 	false    	false   	retrieves the status of an installation                      
dry-run	true     	false   	prints what install would do with the given parameters values
```

cc @itowlson -- do you need this command to output JSON / other formats for the VS Code extension?

### DIFF
--- a/cmd/duffle/bundle.go
+++ b/cmd/duffle/bundle.go
@@ -23,6 +23,7 @@ func newBundleCmd(w io.Writer) *cobra.Command {
 		newBundleSignCmd(w),
 		newBundleVerifyCmd(w),
 		newBundleRemoveCmd(w),
+		newBundleActionsCmd(w),
 	)
 	return cmd
 }
--- a/cmd/duffle/bundle_actions.go
+++ b/cmd/duffle/bundle_actions.go
@@ -1,0 +1,70 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/gosuri/uitable"
+	"github.com/spf13/cobra"
+)
+
+const bundleActionsDesc = ` Lists all actions available in a bundle`
+
+type bundleActionsCmd struct {
+	out        io.Writer
+	bundleFile string
+	insecure   bool
+}
+
+func newBundleActionsCmd(w io.Writer) *cobra.Command {
+	a := &bundleActionsCmd{out: w}
+
+	cmd := &cobra.Command{
+		Use:   "actions BUNDLE",
+		Short: bundleActionsDesc,
+		Long:  bundleActionsDesc,
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			bundle, err := bundleFileOrArg1(args, a.bundleFile)
+			if err != nil {
+				return err
+			}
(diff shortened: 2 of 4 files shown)


itowlson:
JSON output would be useful, thanks - I can parse space-separated lines, but when a value can contain spaces (here the description) it gets a bit messier.

As a heads up, I believe users will find a field named 'stateless' a bit enigmatic; can we find a better name for this?  If not, we should consider flipping it to 'stateful', because 'stateless: false' requires the reader to parse a double negative.  'modifies' is also a bit non-obvious though struggling to think of anything better.

Finally what order guarantees do you give in this command?  Is it the order of the JSON, alphabetical order, the order in which the Go map decides to spit them out this week?  If we display these in a UI we will want a stable order though I can add that in the UI if we don't offer it in the CLI.

radu-matei:
- giving this another thought, the bundle itself is already JSON -- so is this command outputting JSON really helpful after all? It's really not difficult to implement, though.

- I agree that `stateless` and `modifies` are a bit messy -- the decision here was to be consistent with the spec -- that being said, I'm open to changing the display name for these properties -- but there is still some mental gymnastics involved anyway -- you get a value here (let's say for STATEFUL), and you want to change it, then you have to flip the value twice when creating the bundle. If we think other names are more suitable for this, maybe let's have the conversation in the spec?

- this command ranges over a Go map, so there are no guarantees of the order. We could sort the keys and then display them.

itowlson:
Good point about the JSON - I can just do a `duffle show` and parse the JSON directly.

radu-matei:
@michelleN -- updated the UX for bundle reference and file, and added a _tiny_ test.

PR #670, opened 2019-03-19
Add a developer's guide

glyn:


docs/developing.md
```diff
@@ -0,0 +1,43 @@
+# Developer's Guide
+
+## Getting the code
+
+Cloning this repository and change directory to it:
```
silvin-lubecki: You can find the old guide here: https://github.com/deislabs/duffle/commit/3cf48ef2ffb33780ec74267ad3c0af66b18a935f#diff-04c6e90faac2675aa89e2176d2eec7d8L10
I like this Developer's guide 👍 
So maybe you can just get the old sections back to this page?
Note that there's also a `Debugging` page here: https://github.com/deislabs/duffle/blob/master/docs/debugging.md
Maybe you should merge it someway?

glyn: > You can find the old guide here: [3cf48ef#diff-04c6e90faac2675aa89e2176d2eec7d8L10](https://github.com/deislabs/duffle/commit/3cf48ef2ffb33780ec74267ad3c0af66b18a935f#diff-04c6e90faac2675aa89e2176d2eec7d8L10)
> I like this Developer's guide 👍
> So maybe you can just get the old sections back to this page?

Thanks. I already copied everything I thought was worth keeping from that.

> Note that there's also a `Debugging` page here: https://github.com/deislabs/duffle/blob/master/docs/debugging.md
> Maybe you should merge it someway?

I wanted to keep it separate because it is rather specific to vscode users.

radu-matei: Good point, @silvin-lubecki. We decided that the main readme should be user-focused rather than developer-focused.

That being said, when I updated the docs, I overlooked that the info about building the project weren't available somewhere else.

docs/developing.md
```diff
@@ -0,0 +1,43 @@
...
+```console
+$ go get -d github.com/deislabs/duffle/...
+$ cd $(go env GOPATH)/src/github.com/deislabs/duffle
+```
+
+## Building
+
+### Prerequisites
+You need:
+* A working Go 1.11.4 (or later) environment
```
radu-matei: Any particular feature that requires 1.11.4?
I'm running 1.11.1, for example.

michelleN:
@radu-matei - yea we should open an issue and iterate on this 👍

docs/developing.md
glyn: @radu-matei I'm not aware of any. This just gives an impression about the starting point. It would be more relevant if the project was using modules, which it isn't.

Issue #668, opened 2019-03-18
Support image relocation


glyn:
The externals are still to be defined and will be added here in due course.

The intention is to use https://github.com/google/go-containerregistry for accessing OCI image layouts and remote repositories.

silvin-lubecki:
I think the use case you described is exactly what https://github.com/docker/cnab-to-oci does (@simonferquel please jump in if I'm wrong).
You can pull the bundle from a registry, then push it to another (private or not) registry, and `cnab-to-oci` will update the image references. It even support [image promotion pattern](https://docs.docker.com/ee/dtr/user/promotion-policies/internal-promotion/).

With @simonferquel we planned to integrate this library to `duffle` (it's already integrated in `docker-app`), had no time yet to do it.

glyn:
> I think the use case you described is exactly what https://github.com/docker/cnab-to-oci does (@simonferquel please jump in if I'm wrong).
> You can pull the bundle from a registry, then push it to another (private or not) registry, and `cnab-to-oci` will update the image references. It even support [image promotion pattern](https://docs.docker.com/ee/dtr/user/promotion-policies/internal-promotion/).
> 
> With @simonferquel we planned to integrate this library to `duffle` (it's already integrated in `docker-app`), had no time yet to do it.

We have a use case where the bundle is _not_ stored in a registry. Will that also be supported by cnab-to-oci?

silvin-lubecki:
> We have a use case where the bundle is not stored in a registry. Will that also be supported by cnab-to-oci?

I don't understand the use case described earlier then, it says you want to relocate to another registry.
What do you want to relocate then, if it is not towards another registry?
Or maybe you mean you will relocate the images, but not the bundle itself?
That's not how `cnab-to-oci` works. See it more as pushing the thick bundle (bundle + the images) to a registry, as a whole.

glyn:
> Or maybe you mean you will relocate the images, but not the bundle itself?

Correct! So it _seems_ that `cnab-to-oci` won't support our use case. Agreed?

simonferquel:
Yes, the goal of cnab-to-oci is to be able to publish bundles with guaranteed immutability from the bundle metadata down to the actual component and invocation images.
It also guarantees that if you have pull rights for the bundle, you also have pull rights for the invocation image and component images.

silvin-lubecki:
> We have a use case where the bundle is not stored in a registry. 

@glyn could you describe this use case? 
[...]

PR #671, opened 2019-03-20
[WIP] Image relocation

glyn:
This draft PR is a sneak preview for anyone who's interested. Comments gratefully received.


cmd/duffle/relocate.go
```diff
@@ -0,0 +1,155 @@
...
+	//if err != nil {
+	//	return "", nil, err
+	//}
+
+	return "", nil, nil
+}
+
+type FlagsValidator func(cmd *cobra.Command) error
+
+func FlagValidRepository(flagName string) FlagsValidator {
```
radu-matei: nit: is this function used outside of the command? Can it be un-exported?

glyn: No. Yes. Fixed. :-)

Also made `FlagsValidator` private.

cmd/duffle/relocate.go
```diff
@@ -0,0 +1,282 @@
+/*
```
silvin-lubecki: nit: can you clean those copyright headers? I don't think it will fit with duffle licence. @radu-matei ?

glyn: @radu-matei The contributed code is copyright Pivotal and licensed under the Apache License, v2. What else does the duffle license require?

radu-matei:
The linter seems to be unhappy: 

```
golangci-lint run --config ./golangci.yml
cmd/duffle/root.go:50: File is not `goimports`-ed (goimports)
        newRelocateCmd(outLog),
pkg/bundle/bundle.go:68: File is not `goimports`-ed (goimports)
	ImageType      string         `json:"imageType" mapstructure:"imageType"`
	Image          string         `json:"image" mapstructure:"image"`
	OriginalImage  string         `json:"originalImage" mapstructure:"originalImage"`
	Digest         string         `json:"digest,omitempty" mapstructure:"digest"`
	Size           uint64         `json:"size,omitempty" mapstructure:"size"`
	Platform       *ImagePlatform `json:"platform,omitempty" mapstructure:"platform"`
	MediaType      string         `json:"mediaType,omitempty" mapstructure:"mediaType"`
make: *** [lint] Error 1
Makefile:74: recipe for target 'lint' failed
```

cmd/duffle/relocate.go
glyn: Correction. My mistake. The contribution is covered by the MIT license and we are happy to conform to the project standard of not including copyright headers in source files.

glyn:
@radu-matei wrote:
> The linter seems to be unhappy:

Oops. Fixed. Apologies.

glyn:
@radu-matei wrote:
> The linter seems to be unhappy:

Oops. Fixed. Apologies.

glyn:
@radu-matei wrote:
> The linter seems to be unhappy:

Oops. Fixed. Apologies.

pkg/bundle/bundle.go
```diff
@@ -65,12 +65,13 @@ type LocationRef struct {
...
 type BaseImage struct {
-	ImageType string         `json:"imageType" mapstructure:"imageType"`
-	Image     string         `json:"image" mapstructure:"image"`
-	Digest    string         `json:"digest,omitempty" mapstructure:"digest"`
-	Size      uint64         `json:"size,omitempty" mapstructure:"size"`
-	Platform  *ImagePlatform `json:"platform,omitempty" mapstructure:"platform"`
-	MediaType string         `json:"mediaType,omitempty" mapstructure:"mediaType"`
+	ImageType     string         `json:"imageType" mapstructure:"imageType"`
+	Image         string         `json:"image" mapstructure:"image"`
+	OriginalImage string         `json:"originalImage" mapstructure:"originalImage"`
```
radu-matei: In the meantime, we also mergefd #704, which pulls out `pkg/bundle` into a separate directory.

We should make a PR to https://github.com/deislabs/cnab-go with the updated bundle structure, rebase `master` again and pinpoint a version of the Go package.

Let me know if you want me to update `cnab-go` and eventually have a first 0.x release we can use here.

I apologize, I should have spotted this last week.

cmd/duffle/relocate.go
```diff
@@ -0,0 +1,266 @@
...
+
+The --repository-prefix flag determines the repositories for the relocated images.
+Each image is tagged with a name starting with the given prefix and pushed to the repository.
+
+For example, if the repository-prefix is example.com/user, the image istio/proxyv2 is relocated
+to a name starting with example.com/user/ and pushed to a repository hosted by example.com.
+`
+
+type relocateCmd struct {
+	// args
```
radu-matei: We don't usually have comments to separatte args, flags and other things here, but don't really have anything against them -- and might actually start using them in other places as well.

Gopkg.toml
```diff
@@ -53,3 +53,7 @@
 [[constraint]]
   name = "github.com/docker/go"
   version = "1.5.1-1"
+
+[[constraint]]
```
radu-matei: Looking at https://github.com/pivotal/image-relocation -- would it make sense to have an initial release we can pinpoint here?

Not necessarily blocking this PR though.

radu-matei:
Passing an inexistent registry prefix errors out in a rather non-intuitive way:

```
$ duffle relocate cmd/duffle/testdata/relocate/bundle.json relocatedbundle-2 --repository-prefix example.com/user --input-bundle-is-file
Error: unrecognized HTTP status: 404 Not Found
```

silvin-lubecki:
Hello Glynn, as we [discussed](https://github.com/deislabs/duffle/issues/668#issuecomment-475160071) in the issue, I think we should try to improve [cnab-to-oci Fixup](https://github.com/docker/cnab-to-oci/blob/master/remotes/fixup.go) instead.

If I understand it correctly, the only blocking point you saw in `cnab-to-oci` is that we mount/copy all the images in the same repository, and in your case you need to push all the images to their "own" repository?  My guess is that this can be easily implemented in the current Fixup command, it is just a matter of where we push the images, but all the logic of pushing is already there.
@simonferquel and I would be pleased to help you write this PR against cnab-to-oci 👍 

A second argument for not merging this PR as-is, is that it adds a dependency on `github.com/google/go-containerregistry`, which I think will "duplicate" with the `containerd` library used to [push/pull bundles to a registry](https://github.com/deislabs/duffle/pull/681). @radu-matei any thoughts on that point?

glyn:
@silvin-lubecki There are a number of requirements for `duffle relocate` which I’d want to make sure are satisfied by an implementation based on cnab-to-oci:
1. Must preserve content digest of each image during relocation.
2. Must support a one-one repository mapping with similar usability to that provided by the current PR.
3. Must support copying images from OCI image layouts to remote registries so that we can relocate thick bundles.
4. Must be able to relocate a bundle which is _not_ stored in a registry. Isn’t this contrary to the over-arching design of cnab-to-oci?
5. Should have the same authentication configuration as docker CLI.

glyn:
@silvin-lubecki Let's take the discussion over to the issue as I think it's more appropriate [there](https://github.com/deislabs/duffle/issues/668#issuecomment-483595359).

pkg/bundle/bundle.go
glyn: I'll absorb the merge of https://github.com/deislabs/duffle/pull/704 here.

I think we can defer updating to `cnab-go` and pick that up in due course.

Gopkg.toml
glyn: Possibly, although that would make most sense if go-containerregistry was also released, which it hasn't been so far. Let's defer for now.

pkg/bundle/bundle.go
glyn: https://github.com/deislabs/cnab-go/pull/4 raised.

glyn:
https://github.com/deislabs/cnab-go/pull/4 needs merging and we need to upgrade master of this repo to include that before we can resolve the `pkg/bundle/bundle.go` conflict in this PR.

silvin-lubecki:
> My opinion is that we'd rather have this piece of functionality merged, and work to converging the two approaches after, than waiting for it

I'm totally fine with this 👍. Thanks to @glyn we have a clear functional scope, and if it is merged now we can improve `cnab-to-oci` without rushing it.

radu-matei:
Ok, happy we're all back on the same page here.

@glyn -- once the conflicts are fixed, I think this first iteration is good to go.
Thanks a lot!

Gopkg.toml
sajayantony: @glyn  - For helm3 the push/pull utilizes https://github.com/deislabs/oras which is based on containerd @bacongobbler do you think there could be a release of oras for CNAB if needed?

pkg/bundle/bundle.go
glyn: This PR picked up https://github.com/deislabs/cnab-go/pull/4.

Gopkg.toml
glyn: Thanks. The plan is to merge this PR with its implementation in terms of go-containerregistry. The implementation can be changed in the future as appropriate.

glyn:
> > Passing an inexistent registry prefix errors out in a rather non-intuitive way:
> > ```
> > $ duffle relocate cmd/duffle/testdata/relocate/bundle.json relocatedbundle-2 --repository-prefix example.com/user --input-bundle-is-file
> > Error: unrecognized HTTP status: 404 Not Found
> > ```
> 
> I'll address this in the image-relocation repo: [pivotal/image-relocation#6](https://github.com/pivotal/image-relocation/issues/6)

This is now fixed. The error is now:
```
Error: failed to write image example.com/user/deislabs-duffle-50aa5cc4ebb040ac696a9753d1695298@sha256:4d41eeb38fb14266b7c0461ef1ef0b2f8c05f41cd544987a259a9d92cdad2540: unrecognized HTTP status: 404 Not Found
```

cmd/duffle/relocate.go
```diff
@@ -0,0 +1,249 @@
...
+	bun, err := r.setup()
+	if err != nil {
+		return err
+	}
+
+	if err := r.relocate(bun); err != nil {
+		return err
+	}
+
+	if _, err := r.writeBundle(bun); err != nil {
```
silvin-lubecki: nit: The first argument (digest) is dropped here and never used, I guess you can then remove it and only return an error.

cmd/duffle/relocate.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func (r *relocateCmd) relocate(bun *bundle.Bundle) error {
+	// mutate the input bundle to become the output bundle
+	if !r.outputBundleIsFile {
+		bun.Name = r.outputBundle
+	}
+
+	ks := keys(bun.Images)
+	for _, k := range ks {
```
silvin-lubecki: I think it would be simpler to just iterate on the `Images` map and remove this `keys` function:
```go
for k, im := range bun.Images {
    if !isOCI(im.ImageType) && !isDocker(im.ImageType) {
        continue
    }
}
```
You can also "fast return" using a `continue`, but this last point is totally nit picking 👍 .

cmd/duffle/relocate.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+type flagsValidator func(cmd *cobra.Command) error
+
+func flagValidRepository(flagName string) flagsValidator {
+	return func(cmd *cobra.Command) error {
+		repositoryValue := cmd.Flag(flagName).Value.String()
+
+		if strings.HasSuffix(repositoryValue, "/") || strings.Contains(repositoryValue, "//") {
+			return fmt.Errorf("invalid repository: %s", repositoryValue)
```
silvin-lubecki: As you have here a clear error case, maybe you can improve the error message and tell what is the exact issue in the repository name?

cmd/duffle/relocate.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+
+		for i, part := range strings.Split(repositoryValue, "/") {
+			if i != 0 {
+				if strings.ContainsAny(part, ":@\" ") {
+					return fmt.Errorf("invalid repository: %s", repositoryValue)
```
silvin-lubecki: Same here?

cmd/duffle/relocate.go
```diff
@@ -0,0 +1,249 @@
...
+		Short: "relocate images in a CNAB bundle",
+		Long:  relocateDesc,
+		Example: `duffle relocate helloworld hellorelocated --repository-prefix example.com/user
+duffle relocate path/to/bundle.json relocatedbundle --repository-prefix example.com/user --input-bundle-is-file
+duffle relocate helloworld path/to/relocatedbundle.json --repository-prefix example.com/user --output-bundle-is-file`,
+		Args: cobra.ExactArgs(2),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// validate --repository-prefix if it is set, otherwise allow flag omission to be diagnosed as such
+			if cmd.Flags().Changed("repository-prefix") {
+				if err := flagValidRepository("repository-prefix")(cmd); err != nil {
```
silvin-lubecki: Does this function really need to take this string argument and return a functor? As it is only used once, it can be simplified. You can also always check the repository prefix value set:
```go
PreRunE: func(cmd *cobra.Command, args []string) error {
if err := validateRepository(relocate. repoPrefix); err != nil{
    return err
}
```
with bellow, replacing `func flagValidRepository(flagName string) flagsValidator {`:
```go
func validateRepository(repository string) error {
    if strings.HasSuffix(repository, "/") || strings.Contains(repository, "//") {
        return fmt.Errorf(...)
...
```

cmd/duffle/relocate.go
glyn: Fixed.

cmd/duffle/relocate.go
glyn: Fixed.

cmd/duffle/relocate.go
glyn: I was nervous about mutating `bun.Images` while iterating over it, but it seems changing map values during the iteration is safe, so I'm adopting your suggestion.

I'm going to skip introducing the "fast return" because:
* I find the code less easy to read
* any speed improvement would be neglible and performance isn't critical here
* I'd expect the Go compiler's optimiser to strip out unnecessary branches, so the current code is probably just as fast.

cmd/duffle/relocate.go
glyn: Fixed.

cmd/duffle/relocate.go
glyn: I got rid of the functor as suggested. (Functors made more sense in the project where this code originated since there was a lot of flag validation going on.)

Note that `PreRunE` deliberately falls through when the repository prefix is not specified so that we get the cobra error message for a missing required flag (for consistency). I've changed the comment to make this clearer.

glyn:
@silvin-lubecki Thanks for your comments. Changes force pushed.

cmd/duffle/relocate.go
glyn: I looked at the resulting assembly code using `go tool objdump -S -s relocate  duffle` and when `isOCI(im.ImageType) || isDocker(im.ImageType)` is `false`, there is a jump straight back to the map iteration. :-)

silvin-lubecki: Oh sure, but it wasn't an optimisation for performance, but more a coding style. I should have said "early return" by the way, not "fast return", my bad.

cmd/duffle/relocate_test.go
```diff
@@ -0,0 +1,183 @@
...
+	bun, err := loadBundle(bundleFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err = bun.Validate(); err != nil {
+		t.Fatal(err)
+	}
+
+	assertImage := func(t *testing.T, i string, expectedOriginalImageName string, expectedImageName string) {
```
silvin-lubecki: If I read this correctly, the only captured element here is the bundle `bun`. I guess this functor can be extracted as-is as a helper function, by just adding the bundle as a parameter.

glyn: Yes, that's right, only `bun` is captured. However, for consistency, `t` should be captured too. The reason for using a function (not actually a functor, note) here is that it reduces the repetition of arguments at the call sites. I guess it's a matter of personal preference.

PR #673, opened 2019-03-21
Refactoring: move docker driver Simulate field as a config parameter.

silvin-lubecki:
Small refactoring, as I think this Simulate field has nothing to do here, since there's a config map for this kind of parameters.


silvin-lubecki:
@michelleN I don't understand your comment, I already made this [change](https://github.com/deislabs/duffle/pull/673/files#diff-cd53ff31e21090054068e33f6c5672d4R49), is it right?

silvin-lubecki:
PTAL @michelleN @radu-matei

radu-matei:
As the `Simulate` field is only present in the Docker driver, and there is already another way of passing config, I think we should be consistent -- and this approach is the most straightforward.

That being said, also have a look at the Docker driver test below:

```
$ make test
go test  ./...
# github.com/deislabs/duffle/pkg/driver [github.com/deislabs/duffle/pkg/driver.test]
pkg/driver/driver_test.go:58:19: d.(*DockerDriver).Simulate undefined (type *DockerDriver has no field or method Simulate)
```

Thanks!

silvin-lubecki:
Woops, thank you for spotting that @radu-matei ! But that's weird the CI didn't catch it 🤔 
I'll fix that right away 👍

radu-matei:
Curious how none of the two CIs kicked in at all. Investigating..

silvin-lubecki:
Should be fixed + rebased 👍

silvin-lubecki:
PTAL @radu-matei @michelleN

silvin-lubecki:
PTAL @radu-matei @michelleN

PR #674, opened 2019-03-21
Configure container streams

silvin-lubecki:
Add output and error container streams configuration on Docker Driver, different from the Operation.Out, which is there to configure the driver logs.


silvin-lubecki:
PTAL @radu-matei

PR #676, opened 2019-03-21
Adding an install driver using ACI

simongdavies:
This adds a new driver to use Azure ACI to run an invocation image.


pkg/driver/aci_driver.md
```diff
@@ -0,0 +1,145 @@
...
+==> The following new files will be created:
+/home/$USER/.duffle/repositories.json
+==> Generating a new secret keyring at /home/$USER/.duffle/secret.ring
+==> Generating a new signing key with ID $USER <$USER@computer>
+==> Generating a new public keyring at /home/$USER/.duffle/public.ring
+```
+
+3. Install a sample bundle:
+
+A simple helloworld-aci bundle can be found [here]().
```
radu-matei: Missing URL.

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,760 @@
...
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2015-11-01/subscriptions"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/google/uuid"
+
+	"io/ioutil"
```
radu-matei: The linter will not be happy with the ordering of the imports.

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,760 @@
...
+
+	aciName := d.config["ACI_NAME"]
+	d.log("ACI Name:", aciName)
+	if len(aciName) == 0 {
+		aciName = fmt.Sprintf("duffle-%s", uuid.New().String())
+		d.log("ACI Name:", aciName)
+	}
+
+	// TODO ACI does not support file copy
+	// Does not support files because ACI does not support file copy yet
```
radu-matei: Would this be avoided by creating a file share and mounting it?

(Not saying it should be part of this PR, just for future reference.)

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,760 @@
...
+	containerClient := d.getContainerClient()
+	logs, err := containerClient.ListLogs(ctx, aciRG, aciName, aciName, nil)
+	if err != nil {
+		return 0, fmt.Errorf("Error getting container logs :%v", err)
+	}
+
+	lines := strings.Split(strings.TrimSuffix(*logs.Content, "\n"), "\n")
+	noOfLines := len(lines)
+	for currentLine := linesOutput; currentLine < noOfLines; currentLine++ {
+		fmt.Println(lines[currentLine])
```
radu-matei: Should we rather pass the command writer rather than writing directly on stdout?

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,760 @@
...
+	resp, err := containerGroupsClient.Get(ctx, aciRG, aciName)
+	if err != nil {
+		return "", err
+	}
+
+	return *resp.InstanceView.State, nil
+}
+
+func locationIsAvailable(location string, locations []string) bool {
+
```
radu-matei: nit: unnecessary newline here.

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,760 @@
...
+	}
+
+	return false
+}
+func checkForMSIEndpoint() bool {
+	timeout := time.Duration(1 * time.Second)
+	client := http.Client{
+		Timeout: timeout,
+	}
+	_, err := client.Head("http://169.254.169.254/metadata/identity/oauth2/token")
```
radu-matei: nit: could we have the IP address as a constant, and also indicate what it represents?

pkg/driver/aci_driver.go
simongdavies: I don't think that this can be resolved using file mount because the data the needs to be written to the file may be sensitive, also its not possible to mount a file into a pre-existing folder without masking the contents of the folder in the image

pkg/driver/aci_driver.go
simongdavies: I did run make lint

simongdavies:
@radu-matei I have addressed all the comments.

pkg/driver/aci_driver.go
silvin-lubecki: nit: Could you group standard imports together?

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,770 @@
...
+
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+)
+
+const userAgent string = "Duffle ACI Driver"
```
silvin-lubecki: nit: Use a `const` group:
```go
const(
    userAgent = "Duffle ACI Driver"
    msiTokenEndpoint = "http://169.254.169.254/metadata/identity/oauth2/token"
)
```

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,770 @@
...
+const msiTokenEndpoint = "http://169.254.169.254/metadata/identity/oauth2/token"
```
silvin-lubecki: I wonder where this IP comes from? Do we really want this as const in the code? Isn't there any domain name?

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,770 @@
...
+
+// ACIDriver runs Docker and OCI invocation images in ACI
+type ACIDriver struct {
+	config map[string]string
+	// This property is set to true if Duffle is running in cloud shell
+	inCloudShell bool
+	// If true, this will not actually create ACI instance
+	Simulate           bool
```
silvin-lubecki: Why is this field exported ? I don't see where it is used outside this package 🤔

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,770 @@
...
+
+	d.log("Delete Resources:", d.deleteACIResources)
+	d.inCloudShell = len(os.Getenv("ACC_CLOUD")) > 0
+	d.log("In Cloud Shell:", d.inCloudShell)
+	if d.Simulate {
+		return nil
+	}
+
+	err := d.loginToAzure()
+	if err != nil {
```
silvin-lubecki: nit: You can use the following common pattern to check errors:
```go
if err := d.loginToAzure(); err != nil {
    ...
}
```

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,770 @@
...
+		return fmt.Errorf("cannot Login To Azure: %v", err)
```
silvin-lubecki: nit: Why don't you use `%s` here?

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,770 @@
...
+	clientID := d.config["AZURE_CLIENT_ID"]
+	clientSecret := d.config["AZURE_CLIENT_SECRET"]
+	tenantID := d.config["AZURE_TENANT_ID"]
+	applicationID := d.config["AZURE_APP_ID"]
+	d.log("clientID:", clientID)
+	d.log("clientSecret:", clientSecret)
+	d.log("tenantID:", tenantID)
+	d.log("applicationID:", applicationID)
+	// Attempt to login with Service Principal
+	if len(clientID) != 0 && len(clientSecret) != 0 && len(tenantID) != 0 {
```
silvin-lubecki: nit: checking an empty string is more straightforward and simpler to read using the following notation, even if they are equivalent:
```go
if clientID != "" && clientSecret != "" && tenantID != "" {
```

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,770 @@
...
+		if err != nil {
+			return fmt.Errorf("Attempt to login to azure with Service Principal failed: %v", err)
+		}
+
+		d.authorizer = authorizer
+		return nil
+	}
+
+	// Attempt to login with Device Code
+	if len(applicationID) != 0 && len(tenantID) != 0 {
```
silvin-lubecki: This whole function sounds like it could be split into small pieces/sub-functions, with something like:
```go
switch{
case clientID != "" && clientSecret != "" && tenantID != "":
    return loginWithServicePrincipal(...)
case applicationID != "" && tenantID != "":
    return loginWithDeviceCode(...)
case d.inCloudShell:
    return loginWithCloudShell(...)
case checkForMSIEndpoint():
    return loginWithMSI(...)
default:
     return errors.New("Cannot login to Azure - no valid credentials provided")
}
```

pkg/driver/aci_driver.go
```diff
@@ -0,0 +1,770 @@
...
+		d.log("Attempting to Login with Device Code")
+		deviceConfig := auth.NewDeviceFlowConfig(applicationID, tenantID)
+		authorizer, err := deviceConfig.Authorizer()
+		if err != nil {
+			return fmt.Errorf("Attempt to login to azure with Device Code failed: %v", err)
+		}
+
+		fmt.Println("Logged in with Device Code")
```
silvin-lubecki: Shouldn't this be a log instead?

pkg/driver/aci_driver.go
glyn: There is a trend towards using `%v` for errors in case the error has some rich content that isn't expressed well in the string form of the error. I think we should probably agree a consistent approach for duffle and then stick to it...

carolynvs:
@simongdavies I propose that we _not_ attempt to merge the ACI driver in this PR. Duffle supports drivers that are implemented as a separate binary (see [Azure-VM for a POC example](https://github.com/deislabs/duffle/tree/master/drivers/azure-vm)) and ACI driver is the perfect fit for that model.

I have a couple reasons for recommending this path:

* It is specific to a vendor. I'd rather start as we mean to go along and keep it out of tree.
* Duffle is going to be submitted to a standards body very soon, and honestly it's going to be easier for the Azure to work on this driver in a separate repo that Azure can maintain at a candence that they are comfortable with.
* It's a great use case that we can prove out a working solution for other vendors to follow suit and address and problems before duffle 1.0

I am volunteering to help make a repo with Simon's initial commits, setup CI & tests, and collaborate with Simon to get ACI driver working as a standalone driver. What do you think?

carolynvs:
Cool, Simon agreed and we are going to work on this in another repo (https://github.com/deislabs/duffle-aci-driver).

silvin-lubecki:
@carolynvs @simongdavies Should we close this PR then?

simongdavies:
Closing this, I'll work with @carolynvs on this. @glyn @silvin-lubecki Thanks for the feedback , I'll address it in the new version

PR #681, opened 2019-03-27
[WIP] Add push and pull support using cnab-to-oci

radu-matei:
This PR is the very first attempt to make things build. It is a sanity check and a reference for the all dependency versions we need in order to have Duffle and cnab-to-oci compile, and it uses my forks for cnab-to-oci, as well as a CNAB Go package extracted from Duffle in order to avoid circular dependencies.

It also implements naive push and pull commands, to make sure everything works together -- see below.

Obviously, this should _not_ be merged in its current state.

```
radu:deislabs$ duffle push bundle.json -t docker.io/radumatei/duffle-push-test3
Starting to copy image cnab/helloworld:0.1.1...
Completed image cnab/helloworld:0.1.1 copy
WARN[0007] reference for unknown type: application/vnd.cnab.config.v1+json 
Pushed successfully, with digest "sha256:0ad49bc2506ed60a4f8de90703b84fe3c356b1ca0e526fe7200c3d95c4457826"

radu:deislabs$ duffle pull docker.io/radumatei/duffle-push-test3@sha256:0ad49bc2506ed60a4f8de90703b84fe3c356b1ca0e526fe7200c3d95c4457826 -o test3.json
radu:deislabs$ cat test3.json 
{
	"name": "duffle-puhs-test",
	"version": "0.1.1",
	"description": "please, no more go deps shenanigans",
	"keywords": [
		"helloworld",
		"cnab",
		"tutorial"
	],
	"maintainers": [
		{
			"name": "Jane Doe",
			"email": "jane.doe@example.com",
			"url": "https://example.com"
		}
	],
	"invocationImages": [
		{
			"imageType": "docker",
			"image": "radumatei/duffle-push-test3@sha256:a59a4e74d9cc89e4e75dfb2cc7ea5c108e4236ba6231b53081a9e2506d1197b6",
			"size": 942,
			"mediaType": "application/vnd.docker.distribution.manifest.v2+json"
		}
	],
	"images": null,
	"parameters": null,
	"credentials": null
}
```

cc @michelleN, @silvin-lubecki


glyn:
This PR does not appear to satisfy the requirements of https://github.com/deislabs/duffle/issues/668:
* It implements a many-one mapping of repositories; issue 668 requires a one-one mapping.
* It moves a bundle and its images around as an atomic unit; issue 668 relocates just the images of a bundle and does not require either the original or relocated bundle to be stored in a repository.

I've added an agenda item to today's [CNAB community call](https://hackmd.io/s/SyGcBcwQ4) to discuss these points.

radu-matei:
@glyn — this is just a work in progress and reference to get all dependencies working, and as the description states, it will not be merged in its current state.

radu-matei:
@glyn -- I do have one question about this discussion -- and keep in mind that I haven't followed the registry and image relocation discussions closely, so I might be asking something obvious here.

Have we decided that the requirements from #668 about image relocation (which as I recall, is a continuation of the rewrite command) are the also the requirements for the simple bundle push / pull to OCI registries?

At least at first, they look like two separate use cases, both with valid scenarios.

glyn:
> @glyn -- I do have one question about this discussion -- and keep in mind that I haven't followed the registry and image relocation discussions closely, so I might be asking something obvious here.
> 
> Have we decided that the requirements from #668 about image relocation (which as I recall, is a continuation of the rewrite command) are the also the requirements for the simple bundle push / pull to OCI registries?
> 
> At least at first, they look like two separate use cases, both with valid scenarios.

No, we certainly haven't decided that, but I wanted to point out the potential conflict and have the discussion so we are all "on the same page".

radu-matei:
Thanks for pointing this out, @glyn!

To sum up the discussion from the CNAB meeting earlier today, we decided that both approaches are valid, and we want to eventually support relocating the images for `duffle push`, perhaps under a flag.

As we are working on the two implementations and we have a clear view of how they work, we will decide if / how we want to integrate relocating images either directly in Duffle, or as part of the `cnab-to-oci` library used in Duffle for registry support.

Gopkg.toml
```diff
@@ -53,3 +57,46 @@
...
+  source = "github.com/simonferquel/containerd"
+  revision = "42c3614d58767e96c1752977fed94d0b140a0173"
+
+
+[[override]]
+  name = "golang.org/x/sys"
+  revision = "f49334f85ddcf0f08d7fb6dd7363e9e6d6b777eb"
+
+[[override]]
+  name = "k8s.io/apimachinery"
```
silvin-lubecki: Something needs to be cleaned in docker/cli, you shouldn't have to vendor kubernetes just because you want to push and pull...

radu-matei: Yeah, I ended up having to pinpoint this because of `docker/cli`...

silvin-lubecki: I'll try to dig and find the culprit in docker/cli 👍

radu-matei: That would be fantastic!
Let me know how that works, thanks!

radu-matei:
@silvin-lubecki - absolutely, we shouldn't wait on that PR.
Thanks!

Gopkg.toml
radu-matei: @silvin-lubecki  - any updates on solving this in `docker/cli`?
Thanks!

radu-matei:
This PR introduces an initial implementation for working with OCI registries, provided by [`github.com/docker/cnab-to-oci`](github.com/docker/cnab-to-oci).

- `duffle push` - if you logged in to an OCI registry using the Docker CLI, this command will reuse the same credentials, so you should be able to push.

```
$ duffle push test.json --target cnabregistry.azurecr.io/duffle-push-target:latest --bundle-is-file
Starting to copy image ubuntu:latest...
Completed image ubuntu:latest copy
WARN[0002] reference for unknown type: application/vnd.cnab.config.v1+json
Pushed successfully, with digest "sha256:8a6482e6e5778354cbd95cac49ce3cb5fd78ded0df744c889ca3b29cb131f310"
```

As the current implementation uses `cnab-to-oci`, you can [read this WIP document describing the rationale behind `cnab-to-oci`](https://github.com/docker/cnab-to-oci/blob/5295bc728e45e6e293a924dfbec351ea760807a6/README.md).

> Note that you can push a bundle from the bundle store, or a bundle from a file, by passing the `--bundle-is-file` flag.

- `duffle pull` - again, this will reuse the Docker CLI credentials

```
$ rm -rf ~/.duffle/
$ duffle init
==> The following new directories will be created:
/home/radu/.duffle
/home/radu/.duffle/bundles
/home/radu/.duffle/logs
/home/radu/.duffle/plugins
/home/radu/.duffle/claims
/home/radu/.duffle/credentials
==> The following new files will be created:
/home/radu/.duffle/repositories.json
radu:duffle$ duffle pull cnabregistry.azurecr.io/duffle-push-target:latest
radu:duffle$ duffle bundle list
NAME                    VERSION DIGEST
helloworld-test-push    0.1.2   8d100741027eac95a658d354d73bf090f080a318
```

> Note that you can either pull a tag, for example `cnabregistry.azurecr.io/duffle-push-target:latest`, or directly pull the SHA256 - for example `cnabregistry.azurecr.io/duffle-push-target@sha256:8a6482e6e5778354cbd95cac49ce3cb5fd78ded0df744c889ca3b29cb131f310`


This has been tested using Azure Container Registries and Docker Hub.


Feedback needed:

- how the repository is passed for `push`(the `--target` flag) - this PR does not follow the same tag -> push strategy the Docker CLI follows - meaning the *name* of the bundle (as defined by the spec) does not dictate the repository that will be used to push to (as in - you need to first tag an image and include the repository name with the Docker CLI before you can push). In its current form, the bundle will be pushed to whatever `--target` points to, provided you have push access.

- what should Duffle do if the images referenced in the bundle aren't already pushed and available to copy? (@glyn might have some insight here, as he just implemented the `duffle relocate` command`)

Thanks!

cmd/duffle/pull.go
```diff
@@ -0,0 +1,76 @@
...
+	targetRef          string
+	insecureRegistries []string
+	home               home.Home
+}
+
+func newPullCmd(w io.Writer) *cobra.Command {
+	var pull pullCmd
+	cmd := &cobra.Command{
+		Use:   "pull <ref> [options]",
+		Short: "Pulls an image reference",
```
glyn: Does this command really pull an image reference?

glyn:
> * what should Duffle do if the images referenced in the bundle aren't already pushed and available to copy? (@glyn might have some insight here, as he just implemented the `duffle relocate` command`)

I'm not sure what the intended spec of `duffle push` is or should be. This needs to be made clear, preferably in the help text. Specifically:
* Does this command operate on just the bundle itself or on referenced images too?
* If it operates on the reference images, what assumptions does it make about their location and what does it do to them?

(`duffle relocate` only deals with images, but it will read them from their current repositories and write them to their relocated repositories.)

glyn:
> Obviously, this should _not_ be merged in its current state.

Nit: have you heard of _draft_ pull requests? :-)

michelleN:
@glyn Radu had in fact made this a draft pull request and changed it as ready for review in the past few days.

glyn:
> @glyn Radu had in fact made this a draft pull request and changed it as ready for review in the past few days.

Ah, I missed that. Thanks.

cmd/duffle/pull.go
radu-matei: Yeah, thanks for catching this. Updating.

radu-matei:
@michelleN - yes, the `target` flag is required. Adding this as the second argument would look like:

```
duffle push <bundle> <repository>:<optional-tag>
```

That indeed seems a better, updating.

Thanks for the feedback.

radu-matei:
@michelleN - updated the UX with your suggestions, and improved the help messages for both commands.

That error message seems to be tied to some signed bundles (from before pulling out the signing part), and also rebased this, meaning there are no more references to those in this branch, so everything _should_ be fixed now.

cmd/duffle/push.go
```diff
@@ -0,0 +1,112 @@
...
+	insecureRegistries []string
+	allowFallbacks     bool
+}
+
+func newPushCmd(out io.Writer) *cobra.Command {
+	const usage = `Pushes a CNAB bundle to an OCI repository.`
+	const pushDesc = `
+Pushes a CNAB bundle to an OCI registry by pushing all container
+images referenced in the bundle to the target repository (all images are
+pushed to the same repository, and are referenceable through their digest).
```
glyn: Where are images "pushed" _from_? Are they copied direct from their remote repositories or do they have to be present in a docker daemon?

cmd/duffle/pull.go
```diff
@@ -0,0 +1,93 @@
...
+	home               home.Home
+}
+
+func newPullCmd(w io.Writer) *cobra.Command {
+	const usage = `Pulls a bundle from an OCI repository`
+	const pullDesc = `
+Pulls a CNAB bundle from an OCI repository.
+The only argument for this command is the repository where
+the bundle can be found, and by default, this command pulls the
+bundle and stores it in the local bundle store.
```
glyn: Does this command do anything with images referenced by the bundle?

cmd/duffle/push.go
radu-matei: Yes, they are directly copied from the remote (and actually, as #734 tracks, if the images are only present in the local daemon, this will fail)..

cmd/duffle/pull.go
radu-matei: No, by default it does _not_ pull the referenced images.

cmd/duffle/pull.go
```diff
@@ -0,0 +1,93 @@
...
+
+type pullCmd struct {
+	output             string
+	targetRef          string
+	insecureRegistries []string
+	home               home.Home
+}
+
+func newPullCmd(w io.Writer) *cobra.Command {
+	const usage = `Pulls a bundle from an OCI repository`
```
silvin-lubecki: nit: use const block

Gopkg.toml
silvin-lubecki: @radu-matei 2 PR's were merged on docker/cli to remove some dependencies while vendoring the code:
https://github.com/docker/cli/pull/1887 and https://github.com/docker/cli/pull/1890 .
Can you bump docker/cli to master on this PR and check you have less dependencies? 🙏 
By the way docker/cli 19.03 will be release soon, I think we will should pin this version as soon as it is available.

radu-matei: Thanks for the update, will bump th  CLI dependency.

cmd/duffle/pull.go
radu-matei: I’m revisiting this PR after the whole OCI discussion - do you think it pulling the bundle should pull the referenced images?
Most likely they will not be used on the same machine that pulls the bundle, so not sure about this - would an optional flag for pulling the images be helpful?

silvin-lubecki: I think it makes sense for a local development use case to pre-populate the local docker image store (just pull the bundle and everything in it and it "works"), so I agree with the optional flag.

By the way I don't think it should be a `cnab-to-oci` feature, as it has no knowledge of the docker engine, only duffle has. I guess it will then parse the `bundle.json` and call `docker pull` on every docker image using the docker client ?

radu-matei: Yeah, I agree - I wasn't thinking this should be in `cnab-to-oci`.

radu-matei:
Rebased off #827 - haven't addressed all comments yet.

PR #682, opened 2019-03-27
Add a `Merge` method to `credentials.Set`

ijc:
This will merge (union) all keys in a second `Set` into the given one.
Duplicates are not allowed and result in an error.

Signed-off-by: Ian Campbell <ijc@docker.com>

I also debated adding:
```
func (c *CredentialSet) ResolveAndMerge(s Set) (Set, error) {
```
Which would resolve `c` and then `s.Merge()` the values into `s`. WDYT?

### DIFF
--- a/pkg/credentials/credentialset.go
+++ b/pkg/credentials/credentialset.go
@@ -41,6 +41,20 @@ func (s Set) Expand(b *bundle.Bundle, stateless bool) (env, files map[string]str
 	return
 }
 
+// Merge merges a second Set into the base.
+//
+// Duplicate credential names are not allow and will result in an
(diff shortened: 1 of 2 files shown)


ijc:
@radu-matei @michelleN ping? (I think you are the maintainers)

simonferquel:
@michelleN it is something that we use in docker-app here: https://github.com/docker/app/blob/e93616cfbe7db37ebe40af150891347ff4075633/internal/commands/cnab.go#L58.

For a bit of context, as docker-app is opiniated, we have multiple sources of credentials values:
- credentials managed by docker-app (mainly the docker context to use for deploying)
- additional credentials set by the user

PR #687, opened 2019-04-01
Update action.Action interface

silvin-lubecki:
All the action implementations added a 3rd argument `io.Writer`, but this change was never reflected to the actual interface `action.Action`.
I also added static guards to check the actions implements the interface.

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -26,7 +26,7 @@ const stateful = false
 // - status
 type Action interface {
 	// Run an action, and record the status in the given claim
-	Run(*claim.Claim, credentials.Set) error
+	Run(*claim.Claim, credentials.Set, io.Writer) error
 }
 
 func selectInvocationImage(d driver.Driver, c *claim.Claim) (bundle.InvocationImage, error) {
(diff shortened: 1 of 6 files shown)


silvin-lubecki:
The linter is failing on `master` too, here is the PR fixing it https://github.com/deislabs/duffle/pull/688
We should merge the fix before this one.

silvin-lubecki:
@radu-matei I think I don't have access to the logs of the `duffle-ci` check. Can you please take a look?

pkg/action/install.go
```diff
@@ -13,6 +13,9 @@ type Install struct {
 	Driver driver.Driver // Needs to be more than a string
 }
 
+// makes sure Install implements Action interface
+var _ Action = &Install{}
```
radu-matei: I think we used to put this check in `_test.go` files, but I'm not sure this is consistent across the entire code base.

silvin-lubecki: I moved them to their respective test files 👍

silvin-lubecki:
PTAL @radu-matei @michelleN

silvin-lubecki:
Rebased on top of master.

PR #693, opened 2019-04-05
remove signing support from duffle commands

### LINKED ISSUE #696: remove pgp signing support from duffle
The Open-PGP based signing mechanism was removed from the CNAB spec as of [cnab-spec PR 115](https://github.com/deislabs/cnab-spec/blob/master/300-CNAB-security.md). We should remove signing support from all duffle commands to lay the ground work for implementing the [current bundle security section](https://github.com/deislabs/cnab-spec/blob/master/300-CNAB-security.md) of the CNAB spec.

michelleN:
The Open-PGP based signing mechanism was removed from the CNAB spec as of [cnab-spec PR 115](https://github.com/deislabs/cnab-spec/blob/master/300-CNAB-security.md). This PR removes signing support from all duffle commands to lay the ground work for implementing the current [bundle security section](https://github.com/deislabs/cnab-spec/blob/master/300-CNAB-security.md) of the CNAB spec.

resolves #696


cmd/duffle/bundle_show.go
```diff
@@ -25,19 +31,11 @@ func newBundleShowCmd(w io.Writer) *cobra.Command {
 	}
 
 	flags := cmd.Flags()
-	flags.BoolVarP(&bsc.insecure, "insecure", "k", false, "Do not verify the bundle (INSECURE)")
 	flags.BoolVarP(&bsc.raw, "raw", "r", false, "Display the raw bundle manifest")
```
radu-matei: When bundles were signed, `bundle show` would strip the signature, and `--raw` would show the entire file, including the signature -- now that there is no signature anymore, wondering what this flag could be used for -- one thing that comes to mind is to show the canonical vs. pretty form for JSN.

That being said, `bundle show` seems to be broken right now:

```
$ duffle build examples/helloworld
==> Successfully built bundle helloworld:0.1.0
$ duffle  bundle show helloworld:0.1.0
error occurred runtime error: invalid memory address or nil pointer dereference
$ duffle  bundle show helloworld:0.1.0 --raw
{"credentials":null,"description":"A short description of your bundle","images":null,"invocationImages":[{"image":"deislabs/helloworld-cnab:2aa903328e18e41014b0c98c43582169639defa2","imageType":"docker"}],"keywords":["helloworld","cnab","tutorial"],"maintainers":[{"email":"john.doe@example.com","name":"John Doe","url":"https://example.com"},{"email":"jane.doe@example.com","name":"Jane Doe","url":"https://example.com"}],"name":"helloworld","parameters":null,"version":"0.1.0"}
```

cmd/duffle/credential_generate.go
```diff
@@ -53,7 +53,7 @@ func newCredentialGenerateCmd(out io.Writer) *cobra.Command {
 			}
 			csName := args[0]
 
-			bun, err := loadBundle(bf, insecure)
+			bun, err := loadBundle(bf)
```
radu-matei: Is the `insecure` flag definition used anywhere else?

cmd/duffle/bundle_show.go
michelleN: - Yes, I think that's a great idea. Let's make the default, the pretty form and have -r show the canonical form? If it's alright to do that in a follow up PR, I'd like to that instead of adding to this gigantic one lol
- I was getting that error earlier and pushed a fix for it with the last commit. Could you try again and let me know if it's still broken for you?

cmd/duffle/credential_generate.go
michelleN: Good catch. I thought I had gotten all of em but this was the last one! Thanks. Updated.

cmd/duffle/bundle_show.go
radu-matei: Fixed, thanks.

radu-matei: Sure, let's have that in a follow-up commit.

PR #694, opened 2019-04-05
alias install to bundle install

michelleN:
+ update help text.
+ I frequently do a duffle bundle list and then attempt to do a
duffle bundle install and it gives me an error because the command
to use is duffle install. This PR adds the install command under the
bundle cmd group to help with UX/making this easier for humans.


michelleN:
cc/ @carolynvs

PR #697, opened 2019-04-05
update bundle remove flag/help

### LINKED ISSUE #692: duffle bundle remove not working
Errors every time.

michelleN:
resolves #692

### DIFF
--- a/cmd/duffle/bundle_remove.go
+++ b/cmd/duffle/bundle_remove.go
@@ -16,7 +16,7 @@ import (
 const bundleRemoveDesc = `Remove a bundle from the local storage.
 
 This removes a bundle from the local storage so that it will no longer be locally
-available. Bundles can be rebuilt with 'duffle build'.
+available. Bundles can be rebuilt and store in local storage with 'duffle build'.
 
 Ex. $ duffle bundle remove foo  # removes all versions of foo from local store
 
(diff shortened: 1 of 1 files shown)


cmd/duffle/bundle_remove.go
```diff
@@ -16,7 +16,7 @@ import (
 const bundleRemoveDesc = `Remove a bundle from the local storage.
 
 This removes a bundle from the local storage so that it will no longer be locally
-available. Bundles can be rebuilt with 'duffle build'.
+available. Bundles can be rebuilt and store in local storage with 'duffle build'.
```
carolynvs: nit: `and stored in local storage`

cmd/duffle/bundle_remove.go
```diff
@@ -46,7 +46,7 @@ func newBundleRemoveCmd(w io.Writer) *cobra.Command {
 			return remove.run()
 		},
 	}
-	cmd.Flags().StringVar(&remove.versions, "version", "", "A version or SemVer2 version range")
+	cmd.Flags().StringVarP(&remove.versions, "version", "r", "", "A version or SemVer2 version range")
```
carolynvs: So the command should be `duffle bundle rm mybundle -r version` why `-r` as the shortcut for version?

radu-matei: I suppose because `-v` is used for verbose output (at least it's supposed to be used that way, Simon pointed out that it's currently not working as intended).

radu-matei: I guess the alternative would be not to have a shorthand flag at all?

Issue #706, opened 2019-04-10
duffle relocate to relocate invocation images


glyn:
The scope of this issue is to extend [duffle relocate](https://github.com/deislabs/duffle/issues/668) to relocate invocation images (of type oci or docker) as well as other oci/docker images.

A nice to have feature would be the ability to push invocation images to a different registry than the other images.

This feature depends on [spec issue 157](https://github.com/deislabs/cnab-spec/issues/157) which will add `originalImage` and `originalDigest` to invocation images.

See also [duffle relocate for thick bundles based on OCI image layouts](https://github.com/deislabs/duffle/issues/705).

PR #716, opened 2019-04-18
Relocate invocation images

glyn:
Also allow for images not to have digest fields set.

Fixes https://github.com/deislabs/duffle/issues/706

### DIFF
--- a/cmd/duffle/relocate.go
+++ b/cmd/duffle/relocate.go
@@ -22,8 +22,8 @@ import (
 
 const (
 	relocateDesc = `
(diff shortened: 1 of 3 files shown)


radu-matei:
(The only things that changed in master since this PR was opened is the documentation PR, so I think it's safe to merge even if this is out of date.)

glyn:
> (The only things that changed in master since this PR was opened is the documentation PR, so I think it's safe to merge even if this is out of date.)

Agreed, but rebased so that everything is up to date.

PR #723, opened 2019-04-23
ref(pkg/driver): update Lookup method to return error if not supported

vdice:
This refactor proposes to update the `Lookup` method in `driver.go` to return an error if the supplied driver name isn't explicitly supported.


vdice:
(Pushed same/amended commit to try to kick the Azure Pipelines run...)

PR #726, opened 2019-04-24
add CODEOWNERS file

### LINKED ISSUE #710: Add CODEOWNERS file
I think adding a [`CODEOWNERS`](https://help.github.com/en/articles/about-code-owners) file would be really helpful for:

- documenting the maintainers of the project
- automatically requiring PR reviews and blocking the merge unless they are approved, using the branch protection rules.

Looking at the [`OWNERS`](https://github.com/deislabs/cnab-spec/blob/master/OWNERS) file from the spec repo, it seems to be different from the GitHub documented `CODEDOWNERS` format -- @michelleN, is that specific file format used anywhere in particular, or can we update it and use the automatic features from GitHub?

(Here's an [example of how we use this for Brigade](https://github.com/brigadecore/brigade/blob/master/CODEOWNERS)).

Thoughts?

michelleN:
resolves #710

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,4 @@
+# This file is described here:  https://help.github.com/en/articles/about-code-owners
+
+# Global Owners: These members are Core Maintainers of Duffle
+* @technosophos @michellenoorali @vdice @radu-matei @jeremyrickard @carolynvs @bacongobbler


CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# This file is described here:  https://help.github.com/en/articles/about-code-owners
+
+# Global Owners: These members are Core Maintainers of Brigade
```
bacongobbler: typo: brigade should be duffle

michelleN: thank you!

adamreese:
own it

PR #727, opened 2019-04-24
add base governance doc

michelleN:


governance.md
```diff
@@ -0,0 +1,14 @@
+# Governance
+
+## Project Maintainers
+Project maintainers are responsible for activities around maintaining and updating the Duffle, the reference implementation for the [CNAB spec](https://github.com/deislabs/cnab-spec). Final decisions on the project reside with the project maintainers.
```
bridgetkromhout: > updating the Duffle

Do we need "the" here?

governance.md
```diff
@@ -0,0 +1,14 @@
...
+
+Maintainers MUST remain active. If they are unresponsive for >3 months, they will be automatically removed unless a [super-majority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) of the other project maintainers agrees to extend the period to be greater than 3 months.
+
+New maintainers can be added to the project by a [super-majority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) vote of the existing maintainers.
+
+A maintainer may step down by submitting an [issue](https://github.com/deislabs/duffle/issues/new) stating their intent.
```
bridgetkromhout: Is it usual to point to the source of truth for who's a maintainer?

governance.md
michelleN: removed. thank you.

governance.md
michelleN: I added a link to the source of truth but it's in another PR that will be merged in a few mins.

PR #739, opened 2019-05-02
Clarify what duffle is intended for

carolynvs:
Duffle is the reference implementation of the spec. We should point people to the the community tools if they want to build new bundles so that they don't assume that duffle is the official tool for that, and get frustrated when it's super hard. 😀 

We keep a list of community run projects, like Docker Apps and Porter, that we can point to so that people can find a tool to fit their needs. I have an [open PR](https://github.com/deislabs/cnab.io/pull/12) to add Docker Apps to that page, but would like a review from someone at Docker first before I merge it. Maybe @silvin-lubecki can take a look?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,12 +3,15 @@
 [![Waffle.io - Columns and their card count](https://badge.waffle.io/deislabs/duffle.svg?columns=In%20Progress,Needs%20Review,Done)](https://waffle.io/deislabs/duffle)
 
 
-Duffle is the reference implementation of the [CNAB Specification][cnab], a
-command-line tool to install and manage CNAB bundles. While you can build new
-bundles with duffle, we recommend using one of the [CNAB community
-tools][cnab-tools] that were designed to make authoring bundles easier.
+Duffle is the reference implementation of the [CNAB specification][cnab]. It
+provides a comprehensive mapping of _all_ features of the specification, serving
+both as a tool to install and manage bundles, and author bundles at a low level.
 
-To learn more about about CNAB and duffle, check out the [docs](docs/README.md).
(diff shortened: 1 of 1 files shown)


technosophos:
I am not sure I like the way this is phrased. I don't feel like we should say that we _don't recommend_ using Duffle for the things that it is built to be able to do.

It is fine with me if you state that Duffle is a low-level tool designed to show how the specification should work, and that there are other tools that make bundle building easier. But it is not good to imply that one _should not use_ the tool (or even imply that).

To that point, if Duffle is _that bad_ at building bundles, it really should be fixed. (Though the feedback I got was that it was the documentation that was misleading.)

README.md
```diff
@@ -3,7 +3,15 @@
...
 
 
-Duffle is a command line tool that allows you to install and manage CNAB bundles. To learn more about about CNAB and duffle, check out the [docs](docs/README.md).
+Duffle is the reference implementation of the [CNAB specification][cnab]. It
+provides a comprehensive mapping of _all_ features of the specification, serving
+both as a tool to install and manage bundles, and author bundles at a low level.
+
+The community has created implementations of the CNAB spec with
+[opinionated takes on authoring bundles][cnab-tools]. Some even use Duffle's
+[libraries][cnab-go] to handle the CNAB implementation. If you want to make your own CNAB tooling, that is a great place to start!
```
silvin-lubecki: Shouldn't the link `libraries` point to the libraries list on cnab.io, as there are multiple libraries (rust, python, .net) other than go?

carolynvs: Oops! Yes, thank you good catch!

Issue #705, opened 2019-04-10
duffle relocate for thick bundles


glyn:
The scope of this issue is to extend [duffle relocate](https://github.com/deislabs/duffle/issues/668) to relocate oci and docker images from a thick bundle.

Example usage:
```
duffle relocate thickbundle.tgz path/to/relocated.json --input-bundle-is-file --output-bundle-is-file --repository-prefix=example.com/user 
```
Note that the output is a _thin_ bundle: the images will reside in repositories and do not need to be kept in the output bundle.

GreenCee:
One thing I'd like to do is figure out if we would need to separate out the command with a switch namespace or something else that'll indicate this is for OCI images and the associated registry.
I see parallel implementations for VM images and functions or other artifacts

So to fully relocate a fancy thick bundle with all 3 like below it would be

`duffle relocate thickbundle.tgz path/to/relocated.json --input-bundle-is-file --output-bundle-is-file --repository-prefix=example.com/user`
'duffle relocate thickbundle.tgz path/to/relocated.json --input-bundle-is-file --output-bundle-is-file --
vm-blob-container=mydisks'
'duffle relocate thickbundle.tgz path/to/relocated.json --input-bundle-is-file --output-bundle-is-file --lambda-s3-bucket=lambda/functions'

OR

'duffle relocate thickbundle.tgz path/to/relocated.json --input-bundle-is-file --output-bundle-is-file --lambda-s3-bucket=lambda/functions --vm-blob-container=mydisks --repository --prefix=example.com/user'


Does that sound like a reasonable approach? I don't want to explode complexity in branching if it's better to just have 
'duffle relocate oci ...' and 'duffle relocate vm ...'

glyn:
@GreenCee I agree that needs figuring out in due course. Maybe the minimal change required for this PR is to fail the command when an image type other than oci or docker is encountered.

Then when we figure out the way we want to support other image types, probably for a whole set of commands such as import, export, push, pull, relocate, etc., we can implement a backwards compatible change to `duffle relocate` by making the command succeed when it encounters those other image types. 

Does that seem reasonable?

glyn:
> Does that seem reasonable?

I assumed it was.

glyn:
Broadened the scope of this issue to thick bundles containing just oci and docker images in either tar files or an OCI image layout.

glyn:
[...]

PR #740, opened 2019-05-09
duffle relocate of thick bundles

glyn:


glyn:
Closing in favour of the more general https://github.com/deislabs/duffle/pull/740

PR #741, opened 2019-05-09
duffle relocate of thick bundles

glyn:


cmd/duffle/relocate.go
```diff
@@ -2,17 +2,21 @@ package main
 
 import (
 	"fmt"
+	"github.com/deislabs/duffle/pkg/imagestore"
```
radu-matei: Linter will _not_ be happy.

pkg/imagestore/imagestore.go
```diff
@@ -0,0 +1,103 @@
...
+			return nil, err
+		}
+		return tarFilesFactory.ArchiveDir(b.archiveDir).Build()
+	}
+
+	return readOciLayout(layoutDir)
+
+}
+
+func LocatingBuilder() Builder {
```
radu-matei: `NewLocatingBuilder()` maybe?

glyn: Adopted.

cmd/duffle/relocate.go
glyn: Correct. Well spotted.

Gopkg.toml
```diff
@@ -57,7 +57,7 @@
 
 [[constraint]]
   name = "github.com/pivotal/image-relocation"
-  revision = "532dd0b42e7a50010d7868364309cd314a2bb376"
+  revision = "3c9c32bb3d97fc213476f1c9846a99ca35ecba5d"
```
radu-matei: Not blocking this PR, but do you think we could start having a `0.` release for this package?

cmd/duffle/export.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io"
 
+	builder2 "github.com/deislabs/duffle/pkg/imagestore/builder"
```
radu-matei: nit: Why the name `builder2`?

cmd/duffle/relocate.go
```diff
@@ -4,15 +4,22 @@ import (
...
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
 
-	"github.com/pivotal/image-relocation/pkg/image"
-	"github.com/pivotal/image-relocation/pkg/pathmapping"
-	"github.com/pivotal/image-relocation/pkg/registry"
+	"github.com/deislabs/duffle/pkg/imagestore/builder"
+
```
radu-matei: nit: Could you please group the imports from Duffle?

pkg/imagestore/builder.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"io"
+)
+
+// Builder is a means of creating image stores.
+type Builder interface {
+	// ArchiveDir creates a fresh Builder with the given archive directory.
+	ArchiveDir(string) Builder
+
+	// Logs creates a fresh builder with the given log stream.
```
radu-matei: This is rather unintuitive - maybe `NewBuilder`?

pkg/imagestore/builder/builder.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/deislabs/duffle/pkg/imagestore"
+	"github.com/deislabs/duffle/pkg/imagestore/ocilayout"
+	remote2 "github.com/deislabs/duffle/pkg/imagestore/remote"
```
radu-matei: Same as for `builder2` - I guess you didn't want to shadow the `remote` variable, but this is rather counter-intuitive - agian, not something that should block the PR.

pkg/imagestore/ocilayout/ocilayout.go
```diff
@@ -0,0 +1,109 @@
...
+	logs           io.Writer
+}
+
+// ociLayout is an image store which stores images as an OCI image layout.
+type ociLayout struct {
+	layout registry.Layout
+	logs   io.Writer
+}
+
+func NewOciLayout() *Builder {
```
radu-matei: Just curious why you chose not to return an `imagestore.Builder` here.

pkg/packager/export.go
```diff
@@ -7,74 +7,61 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/deislabs/duffle/pkg/imagestore"
+
 	"github.com/deislabs/cnab-go/bundle"
 	"github.com/docker/docker/pkg/archive"
 
 	"github.com/deislabs/duffle/pkg/loader"
```
radu-matei: nit: Could you group the imports from Duffle, please?

cmd/duffle/relocate_test.go
```diff
@@ -1,15 +1,18 @@
...
+	"errors"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
+	"strings"
 	"testing"
 
+	"github.com/deislabs/duffle/pkg/imagestore/imagestoremocks"
+
```
radu-matei: nit: Group imports from Duffle.

radu-matei:
I did another round of reviews for this PR - manually tested and everything seems to work, I do have a couple of comments about the PR:

- could you please also update the description for the `relocate` command (and maybe example) to show it works with thick bundles?

- I have some questions about the general naming and structure for the `imagestore` package.

Thanks!

cmd/duffle/export.go
```diff
@@ -92,20 +94,20 @@ func (ex *exportCmd) run() error {
 }
 
 func (ex *exportCmd) Export(bundlefile string, l loader.BundleLoader) error {
-	is, err := packager.NewImageStore(ex.thin, ex.ociLayout)
+	builder, err := builder2.NewBuilder(ex.thin, ex.ociLayout)
```
silvin-lubecki: nit: I'm not fan at all with the "one letter variable naming" we often see in go, but I think that here I'd prefer see something like this:
```go
b, err := builder.NewBuilder(...)
```
I think that package naming is more important than variable naming.

cmd/duffle/relocate.go
```diff
@@ -97,100 +105,80 @@ duffle relocate helloworld path/to/relocatedbundle.json --repository-prefix exam
...
 
 	return r.writeBundle(bun)
 }
 
-func (r *relocateCmd) relocate(bun *bundle.Bundle) error {
-	// mutate the input bundle to become the output bundle
-	if !r.outputBundleIsFile {
-		bun.Name = r.outputBundle
+// The caller is responsible for deleting the returned temporary directory, which may contain the returned bundle.
+func (r *relocateCmd) setup() (*relocator.Relocator, *bundle.Bundle, string, error) {
```
silvin-lubecki: You could also return a "cleaning" functor, so the caller has only to defer the call of this function, without having to know what is the cleaning logic.

Gopkg.toml
glyn: https://github.com/pivotal/image-relocation/releases/tag/v0.1

cmd/duffle/export.go
glyn: That was unintentional and due to a refactoring without sufficient eye-balling afterwards. Will fix.

glyn: Actually, it has already gone!

cmd/duffle/relocate.go
glyn: Fixed.

pkg/imagestore/builder.go
glyn: Maybe the description is confusing, but this is just the builder pattern which allows chaining. It doesn't mutate the "receiver" but rather returns a fresh builder with the logs value set. Not sure how to explain this better in the code without linking to the builder pattern somewhere...

pkg/imagestore/builder/builder.go
glyn: Same issue again - I refactored a variable and didn't pay enough attention afterwards. Fixed.

pkg/imagestore/builder.go
radu-matei: Ok,I thought that might be the case - maybe consider a more intuitive naming for the method, like `SetLogs`?

The only reservation I have around the naming for this is the fact that we have an additional `Builder` interface which also has a `Logs()` method, but that one returns the logs from the build - which might generate confusion.
But if you have strong opinions about that, we can also have a look and consider refactoring later.

pkg/imagestore/ocilayout/ocilayout.go
glyn: That's mainly because it's idiomatic Go to return concrete types and leave any association with interfaces to the callers. Sometimes this approach runs out of steam as I think it did elsewhere in this PR and you _have_ to return an interface type in order for Go to match methods (typically in another interface), but generally returning concrete types is normal practice.

There are several articles on this topic, of which http://idiomaticgo.com/post/best-practice/accept-interfaces-return-structs/ isn't bad.

pkg/packager/export.go
glyn: Done. I must admit that I hardly ever read imports blocks so I'm glad you pointed this out.

cmd/duffle/relocate_test.go
glyn: Done.

pkg/imagestore/ocilayout/ocilayout.go
glyn: BTW I came to Go from Java and I really didn't like this idiom when I first came across it, but I've since grown to accept it and even see the benefits. The only downside from my perspective is that it doesn't capture _intention_ quite so well in some situations.

glyn:
> I did another round of reviews for this PR - manually tested and everything seems to work, I do have a couple of comments about the PR:
> 
> * could you please also update the description for the `relocate` command (and maybe example) to show it works with thick bundles?

An excellent point and something I normally try to do early on, so thanks for spotting this omission! Add both more description and an example.

> * I have some questions about the general naming and structure for the `imagestore` package.
> 
> Thanks!

I'm all ears...

cmd/duffle/export.go
glyn: Absolutely. This crept in due to refactoring without sufficient eye-balling afterwards. Fixed.

cmd/duffle/relocate.go
glyn: Great idea. Done.

nit: it was merely a function, not a functor.

glyn:
@radu-matei @silvin-lubecki I've addressed all issues which you raised and re-based on `master`. Please could you approve and/or raise further issues at your earliest convenience.

glyn:
Closing this PR as there are a couple of spec issues which will require rework to it:
* https://github.com/deislabs/cnab-spec/issues/195
* https://github.com/deislabs/cnab-spec/issues/182

PR #743, opened 2019-05-10
Improve panic handling

glyn:
During development, panics with values which are not errors cause duffle to
exit silently and with exit code 0.

For instance:

panic("not yet implemented")

results in a silent exit.

Also, panics which are errors only result in the error being printed rather
than a stack trace.

It is better not to recover from the panic and allow the standard diagnostics
to be printed. This shouldn't affect normal UX since panics indicate a
programming error.

### DIFF
--- a/cmd/duffle/main.go
+++ b/cmd/duffle/main.go
@@ -27,14 +27,6 @@ var (
 )
 
 func main() {
-	defer func() {
-		if r := recover(); r != nil {
-			if err, ok := r.(error); ok {
-				fmt.Fprint(os.Stderr, fmt.Sprintf("error occurred %+v", err))
-				must(err)
-			}
-		}
-	}()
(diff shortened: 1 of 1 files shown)


radu-matei:
I mostly agree with this, but I'll also let @carolynvs comment on this.

Thanks!

glyn:
I was interested in @technosophos's review too since I think he introduced the recovery code.

technosophos:
I do not believe that I did initially introduce the panic handler. It has been there since `main()` was first added. https://github.com/deislabs/duffle/commit/94867a0dbd1928a8dfdae9126540dcf0192d74d2#diff-7e150ed5edcb6baedce4c675ee74abd8

That said, I don't have a problem removing it. It was likely just a protective measure for preventing panics during the very early days of Duffle dev.

PR #746, opened 2019-05-10
Fix constraint for cnab-go to point to origin

radu-matei:
#742 included the branch of `deislabs/cnab-go` from the PR, which should have been updated before merging.. This PR updates the constraint to point to a revision of origin, rather than the PR.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -65,9 +65,8 @@
 
 [[constraint]]
   name = "github.com/deislabs/cnab-go"
-  source = "github.com/radu-matei/cnab-go"
-  branch = "add-duffle-types"
-
+  revision = "9028f0f800a40977c2e49a38be764978cfe82324"
+  
 [[override]]
   name = "k8s.io/apimachinery"
   revision = "kubernetes-1.11.2"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -180,7 +180,6 @@
   version = "v1.1.1"
 
 [[projects]]
-  branch = "add-duffle-types"
   digest = "1:26492e179ef4aebabd64482e08144f77273a817f0ad865021d4c7172d78c9892"
   name = "github.com/deislabs/cnab-go"
   packages = [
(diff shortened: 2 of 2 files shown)


Gopkg.toml
```diff
@@ -65,9 +65,8 @@
 
 [[constraint]]
   name = "github.com/deislabs/cnab-go"
-  source = "github.com/radu-matei/cnab-go"
-  branch = "add-duffle-types"
-
+  revision = "9028f0f800a40977c2e49a38be764978cfe82324"
```
carolynvs: I'd much prefer to see us tag a version number on cnab-go and then change this to `version = "0.1.0"`.

radu-matei: I tried to avoid adding a version, since it would introduce yet another one that is not the same with the spec or Duffle...

radu-matei: Ok, what tag version number do you suggest for now?

The only thing I don't really know how to avoid is confusion from introducing yet another version number, not in sync with the spec or Duffle..

carolynvs: Pinning to a specific version causes problems with dep's solving algorithm which is why I am so strongly saying don't do that.

When in doubt, use the semver, Radu. If your plan was to use the exact same version # between duffle, cnab-spec and cnab-go (not sure I agree) then stick with what we are doing with the other tools and version it with `0.1.0-ralpha.1+englishrose`. Or if you want to be forward looking, `1.0.0-alpha1` and just keep bumping that last # until the cnab spec is released at v1, then after that you can use the minor and patch since cnab doesn't use those values and only has a major.

Other strategies:

* `0.1.0+cnab1` lets you add build metadata indicating compatibility with cnab v1 without triggering prerelease handling.
* Dual tagging like k8s client does, where you have a version number that makes sense for your own client, and then non semver tags to indicate compatibility with cnab spec versions.

jeremyrickard:
We've already updated duffle to use cnab-go tagged releases, so I'm going to close this.

Issue #745, opened 2019-05-10
Turn off automatic review requests


glyn:
Currently a bot automatically requests a review from all the core maintainers.

I’m not keen on a bot doing this as, with a large set of maintainers, it clouds what’s really required.

In other projects, maintainers subscribe to the repo and keep an eye on PRs.

Similarly, raisers of PRs should be free to nominate particular reviewers (oh except they can’t because I don’t think we have any automation like prow to allow them to do that, but at least they can add a comment `/cc @somemaintainer`).

silvin-lubecki:
Could we also prevent merging a PR onto master without at least 2 approvals?

glyn:
Yes: 
![Screenshot 2019-05-10 at 15 26 47](https://user-images.githubusercontent.com/94953/57534517-176ea200-7338-11e9-9965-03ac07c6def6.png)

carolynvs:
I'm not clear if duffle needs 1 or 2 reviewers? I know cnab-spec needs 2.

glyn:
> I'm not clear if duffle needs 1 or 2 reviewers? I know cnab-spec needs 2.

That was my recollection for cnab-spec, but it's not currently enforced.

carolynvs:
I just went rogue and turned on enforcing the CLA bot, and the validation test and requiring 2 reviews on cnab-spec. I don't feel comfortable changing the duffle settings since we don't have clear recollection/agreement on the rules to enforce though.

carolynvs:
Also I too would really like the bot turned off and instead have people be able to assign reviewers via a comment like prow. When everyone is assigned, no one feels responsible, and it's hard to track which PRs you really cared about reviewing.

technosophos:
Yeah, we need to require two approvals per PR, as @silvin-lubecki suggests.

carolynvs:
I have fixed our GH settings to require 2 reviews just like cnab-spec.

silvin-lubecki:
Should we close the issue then ? Or is the bot asking for reviewers still up?

carolynvs:
I didn't change the bot because I have no clue where it's configured. 😀 So yeah we need to leave this open. Maybe @radu-matei knows?

technosophos:
The "bot" is actually just the way that CODEOWNERS works. Anyone on the `*` line will be added to the Reviewers list, since in GitHub's opinion, this is what a global owner does.

[...]

PR #755, opened 2019-05-15
Limit automatic assignment of PR reviewers

glyn:
Fixes https://github.com/deislabs/duffle/issues/745


technosophos:
meh. Honestly I have mixed feelings about this. I feel like you turned "I don't like notifications" into "nobody gets added to every ticket", and that's a process change. I'd be open to giving it a try, though.

technosophos:
WHOA THERE... why the fast merge?

PR #756, opened 2019-05-15
update to make output from command driver available immediately

simongdavies:
This change updates the commanddriver so that when a driver writes to stdout or stderr it is sent to duffle immediately rather than blocking until the driver completes

### DIFF
--- a/pkg/driver/command_driver.go
+++ b/pkg/driver/command_driver.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"strings"
@@ -71,7 +72,34 @@ func (d *CommandDriver) exec(op *driver.Operation) error {
 	}
 	cmd.Env = pairs
 	cmd.Stdin = bytes.NewBuffer(data)
-	out, err := cmd.CombinedOutput()
-	fmt.Fprintln(op.Out, string(out))
-	return err
+
+	// Make stdout and stderr from driver available immediately
+
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return fmt.Errorf("Setting up output handling for driver (%s) failed: %v", d.Name, err)
+	}
+
+	go func() {
+
+		// Errors not handled here as they only prevent output from the driver being shown, errors in the command execution are handled when command is executed
+
+		io.Copy(op.Out, stdout)
+	}()
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		return fmt.Errorf("Setting up error output handling for driver (%s) failed: %v", d.Name, err)
+	}
+	go func() {
+
+		// Errors not handled here as they only prevent output from the driver being shown, errors in the command execution are handled when command is executed
+
+		io.Copy(op.Out, stderr)
+	}()
+
+	if err = cmd.Start(); err != nil {
+		return fmt.Errorf("Start of driver (%s) failed: %v", d.Name, err)
+	}
+
+	return cmd.Wait()
 }


pkg/driver/command_driver.go
```diff
@@ -71,7 +72,26 @@ func (d *CommandDriver) exec(op *driver.Operation) error {
 	}
 	cmd.Env = pairs
 	cmd.Stdin = bytes.NewBuffer(data)
-	out, err := cmd.CombinedOutput()
-	fmt.Fprintln(op.Out, string(out))
-	return err
+
+	// Make stdout and stderr from driver available immediately
+
+	stdout, _ := cmd.StdoutPipe()
```
glyn: The error should be handled.

pkg/driver/command_driver.go
```diff
@@ -71,7 +72,26 @@ func (d *CommandDriver) exec(op *driver.Operation) error {
...
+	go func() {
+		io.Copy(op.Out, stdout)
+	}()
+	stderr, _ := cmd.StderrPipe()
```
glyn: The error should be handled.

pkg/driver/command_driver.go
```diff
@@ -71,7 +72,26 @@ func (d *CommandDriver) exec(op *driver.Operation) error {
...
+	go func() {
+		io.Copy(op.Out, stderr)
```
glyn: This is a matter of taste, but did you consider handling the error?

pkg/driver/command_driver.go
```diff
@@ -71,7 +72,26 @@ func (d *CommandDriver) exec(op *driver.Operation) error {
...
+	}()
+	err = cmd.Start()
+	if err != nil {
```
glyn: This project favours the style `if err := cmd.Start; err != nil {` for cases where no output variable need leaving in scope.

pkg/driver/command_driver.go
```diff
@@ -71,7 +72,26 @@ func (d *CommandDriver) exec(op *driver.Operation) error {
...
+		return fmt.Errorf("cmd.Start(%s) failed: %v", d.cliName(), err)
```
glyn: nit: the error message will make more sense to the user if it is phrased in terms of externals the user will understand rather than internal code details. So something like "start of command driver %s failed: %v", although I don't understand the context well enough to provide a good message.

pkg/driver/command_driver.go
```diff
@@ -71,7 +72,26 @@ func (d *CommandDriver) exec(op *driver.Operation) error {
...
+	}
+
+	err = cmd.Wait()
+	if err != nil {
```
glyn: Fold these two lines together like the example earlier.

pkg/driver/command_driver.go
```diff
@@ -71,7 +72,26 @@ func (d *CommandDriver) exec(op *driver.Operation) error {
...
+		return err
+	}
+
+	return nil
```
glyn: In fact, you can contract the last few lines into the following if you like:
```
return cmd.Wait()
```
Not everybody likes this style though - take your pick.

pkg/driver/command_driver.go
```diff
@@ -71,7 +72,26 @@ func (d *CommandDriver) exec(op *driver.Operation) error {
...
 	cmd.Stdin = bytes.NewBuffer(data)
-	out, err := cmd.CombinedOutput()
-	fmt.Fprintln(op.Out, string(out))
-	return err
+
+	// Make stdout and stderr from driver available immediately
+
+	stdout, _ := cmd.StdoutPipe()
+	go func() {
+		io.Copy(op.Out, stdout)
```
glyn: This is a matter of taste, but did you consider handling the error?

pkg/driver/command_driver.go
simongdavies: @glyn I didnt handle the error here intentionally as I dont think we should fail the command because of a failure to produce output

pkg/driver/command_driver.go
simongdavies: @glyn as with the stderr case I didnt handle the error here intentionally as I dont think we should fail the command because of a failure to produce output

pkg/driver/command_driver.go
glyn: Fine. Might be worth saying that in a comment?

simongdavies: Done !!

vdice:
Just chiming in here to certify the commit in this PR passes CI locally.  With the LGTMs above, I'll therefore merge.

PR #765, opened 2019-06-10
Add a kubernetes driver

### LINKED ISSUE #763: Implement kubernetes driver
A kubernetes driver would run the invocation image in kubernetes, probably as some kind of job with cluster admin privileges.

Note that the image would not have access to a docker daemon.

This should provide a workaround for the problem of needing to patch `~/.kube/config` for Docker Desktop. See https://github.com/deislabs/duffle/issues/762 for context.

jlegrone:
Closes #763 

This PR includes new integration tests that are behind a build flag. To run them, you'll need to be authenticated to a kubernetes cluster with permissions in the `default` namespace. Then run `go test` with the integration tag:

```
go test -tags=integration ./...
```


jlegrone:
cc @technosophos -- I know you've done some work towards this in #560

pkg/driver/kubernetes_driver_test.go
```diff
@@ -0,0 +1,47 @@
+package driver
+
+// TODO(jlegrone): uncomment tests when k8s.io/client-go/kubernetes/fake is available
```
jlegrone: We need to update k8s.io/client-go in order to get the fake client implementations.

pkg/driver/kubernetes_driver.go
```diff
@@ -0,0 +1,366 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const (
+	k8sContainerName    = "invocation"
+	k8sFileSecretVolume = "files"
+)
+
+// KubernetesDriver is capable of running an invocation image in a Kubernetes cluster.
```
glyn:
```suggestion
// KubernetesDriver runs an invocation image in a Kubernetes cluster.
```

pkg/driver/kubernetes_driver.go
```diff
@@ -0,0 +1,366 @@
...
+	driver.setDefaults()
+	err := driver.setClient(conf)
+	if err != nil {
+		return driver, err
+	}
+
+	return driver, nil
+}
+
+// Config returns the Kubernetes driver configuration options
```
glyn:
```suggestion
// Config returns the Kubernetes driver configuration options.
```

pkg/driver/kubernetes_driver.go
```diff
@@ -0,0 +1,366 @@
...
+func (k *KubernetesDriver) Config() map[string]string {
+	return map[string]string{
+		"KUBE_NAMESPACE":  "Increase verbosity. true, false are supported values",
```
glyn: Values doesn't seem to match key.

pkg/driver/kubernetes_driver.go
```diff
@@ -0,0 +1,366 @@
...
+		"SERVICE_ACCOUNT": "Always pull image, even if locally available (0|1)",
```
glyn: Values doesn't seem to match key.

pkg/driver/kubernetes_driver.go
```diff
@@ -0,0 +1,366 @@
...
+		"KUBE_CONFIG":     "Absolute path to the kubeconfig file",
+		"MASTER_URL":      "Kubernetes master endpoint",
+	}
+}
+
+// SetConfig sets Kubernetes driver configuration
```
glyn:
```suggestion
// SetConfig sets Kubernetes driver configuration.
```

pkg/driver/kubernetes_driver.go
```diff
@@ -0,0 +1,366 @@
...
+	k.BackoffLimit = 0
+	k.ActiveDeadlineSeconds = 300
+	k.requiredCompletions = 1
+	k.deletionPolicy = metav1.DeletePropagationBackground
+}
+
+func (k *KubernetesDriver) setClient(conf *rest.Config) error {
+	coreClient, err := coreclientv1.NewForConfig(conf)
+	if err != nil {
+		return err
```
glyn: Why return an error to a caller which panics? "First failure data capture" would be better by panicking here. Alternatively, errors could be returned everywhere after logging a suitable message and then panic just before the driver loses control (unless the caller of the driver panics if an error is returned). The latter approach is idiomatic Go and lends itself to easier unit testing.

pkg/driver/kubernetes_driver.go
```diff
@@ -0,0 +1,366 @@
...
+		secret := &v1.Secret{
+			ObjectMeta: meta,
+			StringData: op.Environment,
+		}
+		secret.ObjectMeta.GenerateName += "env-"
+		envsecret, err := k.secrets.Create(secret)
+		if err != nil {
+			return err
+		}
+		// TODO: add to slice of resources awaiting cleanup instead, then return err
```
glyn: Looks like a TODO was left over. Did you intend to address it in this PR?

pkg/driver/kubernetes_driver.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+	k.jobs = batchClient.Jobs(k.Namespace)
+	k.secrets = coreClient.Secrets(k.Namespace)
+	k.pods = coreClient.Pods(k.Namespace)
+
+	return nil
+}
+
+// Run executes the operation inside of the invocation image
+func (k *KubernetesDriver) Run(op *driver.Operation) error {
```
glyn: Personally, I think this method is too long and would benefit from factoring out helper methods/functions.

pkg/driver/kubernetes_driver.go
```diff
@@ -0,0 +1,366 @@
...
+
+	go func() {
+		watching := map[string]bool{}
+		for event := range watch.ResultChan() {
+			pod, ok := event.Object.(*v1.Pod)
+			if !ok {
+				continue
+			}
+			podName := pod.GetName()
+			if watching[podName] {
```
glyn: I find this a little confusing because `watching` seems to mean "already watched". Perhaps rename the variable?

pkg/driver/kubernetes_driver.go
```diff
@@ -0,0 +1,366 @@
...
+	})
+}
+
+func (k *KubernetesDriver) deleteJob(name string) error {
+	return k.jobs.Delete(name, &metav1.DeleteOptions{
+		PropagationPolicy: &k.deletionPolicy,
+	})
+}
+
+// TODO: limit character count
```
glyn: Was this meant to be done in this PR?

pkg/driver/kubernetes_driver_test.go
glyn: I think you should generate a fake (somehow) and check it in to this repo (possibly as a temporary measure until the above fake client is available) so that a good unit test case be provided in this PR. Otherwise, it's likely that the code under test won't be properly structured for unit testing and will therefore have poor coverage.

pkg/driver/kubernetes_driver.go
jlegrone: I just double checked, and it looks like this gets automatically truncated by the API server.

pkg/driver/kubernetes_driver.go
jlegrone: Actually I think this is fine. Removed the comment.

pkg/driver/kubernetes_driver.go
jlegrone: There are actually two callers -- `NewKubernetesDriver` does return the error:

https://github.com/deislabs/duffle/blob/206a2bec45ddddb9dcf558581f27249a80db835f/pkg/driver/kubernetes.go#L45-L53

The `Configurable` interface's `SetConfig` method signature doesn't have an error return, so I figured panicking was the best thing to do for now. A better solution would probably be to have `SetConfig` return an error since any operation accepting user input is likely to be fallible anyway.

pkg/driver/kubernetes_driver.go
glyn: Still long, but that shouldn't block this PR.

pkg/driver/kubernetes.go
```diff
@@ -0,0 +1,372 @@
...
+		return fmt.Errorf("KUBE_NAMESPACE is required")
+	}
+	labelMap := generateLabels(op)
+	meta := metav1.ObjectMeta{
+		Namespace:    k.Namespace,
+		GenerateName: generateNameTemplate(op),
+		Labels:       labelMap,
+	}
+	// Mount SA token if a non-zero value for ServiceAccountName has been specified
+	mountServiceAccountToken := k.ServiceAccountName != ""
```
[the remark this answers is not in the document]
jlegrone: This was an intentional decision, as we don't know that the kubernetes driver is being used to run invocation images that deploy to the same cluster, or even to kubernetes. I would like to avoid giving permissions to invocation images that haven't been explicitly declared as required by the bundle.

That said, duffle could afford to be more opinionated and set a default service account here:
https://github.com/deislabs/duffle/blob/206a2bec45ddddb9dcf558581f27249a80db835f/pkg/driver/lookup.go#L14-L15

Thoughts?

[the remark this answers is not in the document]
jlegrone: @sbawaska also did you try mounting kubeconfig as a credential as requested by the riff bundle?

pkg/driver/kubernetes.go
```diff
@@ -0,0 +1,372 @@
...
+	k.secrets = coreClient.Secrets(k.Namespace)
+	k.pods = coreClient.Pods(k.Namespace)
+
+	return nil
+}
+
+// Run executes the operation inside of the invocation image.
+func (k *KubernetesDriver) Run(op *driver.Operation) error {
+	if k.Namespace == "" {
+		return fmt.Errorf("KUBE_NAMESPACE is required")
```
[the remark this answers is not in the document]
jlegrone: Similar idea as https://github.com/deislabs/duffle/pull/765#discussion_r292991827. Credentials mounted in invocation images are readable by anyone with secret `list` permissions in the target namespace. In a multi-tenant cluster using the `default` namespace would really inappropriate IMO, so from the driver's perspective I think it's best to require a value to be set explicitly.

Installer implementations like duffle may still set a default namespace as this is a public field.

pkg/driver/kubernetes.go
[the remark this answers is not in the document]
sbawaska: I think the missing piece here is a lack of contract between the driver and the bundle for the parameters needed by the driver.

To me, the use case for this driver is to run the invocationImage in cluster where the bundle is to be installed. I think we should let the default docker driver handle the deploying to non-kubernetes environments. 

I like this quote from Allan Kay: 
> Simple things should be simple, complex things should be possible.

In this context it would mean: provide defaults, but allow the bundle author to override the defaults. more concretely, I think we should:
1. provide `default` as the default value for SERVICE_ACCOUNT. The bundle author could set this as a parameter and optionally provide a default value of their own in bundle.json.
2. provide default namespace from `/var/run/secrets/kubernetes.io/serviceaccount/namespace` for KUBE_NAMESPACE, again the bundle author could set this as a parameter too.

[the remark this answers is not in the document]
sbawaska: On second though, we could iterate on these issues to make the experience better, and work on some contract between driver and duffle for parameters needed. Let's merge this!

pkg/driver/kubernetes.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+	labelMap := generateLabels(op)
+	meta := metav1.ObjectMeta{
+		Namespace:    k.Namespace,
+		GenerateName: generateNameTemplate(op),
+		Labels:       labelMap,
+	}
+	// Mount SA token if a non-zero value for ServiceAccountName has been specified
+	mountServiceAccountToken := k.ServiceAccountName != ""
+	job := &batchv1.Job{
```
glyn: @sbawaska demonstrated this PR to our team today and it was very well received. However, it was pointed out that in some circumstances jobs can restart even when you think you have configured them not to. I raised https://github.com/deislabs/duffle/issues/771 to cover this. Feel free to implement this issue in this PR or merge and we can implement it later.

glyn: Actually, upon further investigation, it seems that the situation when jobs restart when we don't want them to may have been fixed in k8s, so I've closed https://github.com/deislabs/duffle/issues/771. Sorry for the false alarm.

pkg/driver/kubernetes.go
```diff
@@ -0,0 +1,372 @@
...
+	var kubeconfig string
+	if kpath := settings["KUBE_CONFIG"]; kpath != "" {
+		kubeconfig = kpath
+	} else if home := homeDir(); home != "" {
+		kubeconfig = filepath.Join(home, ".kube", "config")
+	}
+
+	conf, err := clientcmd.BuildConfigFromFlags(settings["MASTER_URL"], kubeconfig)
+	if err != nil {
+		panic(err)
```
silvin-lubecki: `KEEP CALM AND DON'T PANIC` 😺 
Wouldn't it be better to modify the `driver.Configurable` interface so `SetConfig` can return an error?

jlegrone: I actually did exactly this and then realized the interface is defined in cnab-go. I'm definitely in favor of this, just has to be a separate pr.

pkg/driver/kubernetes_driver.go
jeremyrickard: Let's open an issue in cnab-go for this.

pkg/driver/kubernetes.go
technosophos: Yeah, this can be addressed in a follow-on PR when we can fix the interface.

silvin-lubecki: Sure I'm fine with a follow-up 👍

PR #766, opened 2019-06-11
build / release makeover

krancour:
By request from @jeremyrickard 

Highlights:

- Tracks vendored code (it's a modest 36 MB or so)
- All dev tasks (tests, lint checks, etc) occur within a consistent, containerized environment
- Multi-stage image builds
- DinD image builds (don't require the host's docker socket to be mounted into containers)

cc @vdice


bacongobbler:
Isn't the purpose of a lockfile to pin dependencies at particular versions which can rebuild the vendor tree? I'm not following why the vendor directory needs to be checked into source control.

docs/developing.md
```diff
@@ -9,46 +9,94 @@ $ cd $(go env GOPATH)/src/github.com/deislabs/duffle
 ```
 
 ### Prerequisites
+
 You need:
-* A working Go 1.11.4 (or later) environment
```
glyn: If I run `make build-all-bins` without `go` in the path, the command fails with:
```
bash -c "LDFLAGS=\"-w -s -X github.com/deislabs/duffle/pkg/version.Version=c6a43ae\" scripts/build.sh"
building linux-amd64
scripts/build.sh: line 11: go: command not found
make: *** [build-all-bins] Error 127
```
So it seems that go is still needed.

docs/developing.md
```diff
@@ -9,46 +9,94 @@ $ cd $(go env GOPATH)/src/github.com/deislabs/duffle
...
+
+```console
+$ make build-all-bins
+```
+
+To build for one specific OS / architecture:
+
 ```console
-$ sudo make install
+$ OS=<desired OS> ARCH=<desired architecture> make build-bin
```
glyn: This is part of my local development cycle, so would it be possible to add a simple alias such as `make duffle` which would produce `bin/duffle`?

docs/developing.md
krancour: You may have caught a mistake here. Will look.

docs/developing.md
krancour: I struggled with what to do about that target. It doesn't work on Windows...

krancour:
> I'd like a quick/memorable way to make bin/duffle on macOS or linux.

See doc changes. There is a way, but perhaps not adequately quick or memorable. I would be happy to add a few OS-specific convenience targets.

krancour:
> Isn't the purpose of a lockfile to pin dependencies at particular versions which can rebuild the vendor tree? I'm not following why the vendor directory needs to be checked into source control.

This is a classic debate and from what I have observed, `dep` users have mostly been erring on the side of tracking vendored code for some time now. You are absolutely correct that the .lock for enables a dependency tree to be rebuilt... as long as no one has deleted any library from github...

Beyond that, when vendor code isn't tracked, there's a need for some additional dependency resolution step (e.g. `make bootstrap`) to be run. For good measure, that should be run every time you check out a given branch, else you're potentially building with the wrong dependencies. So tracking vendored code has the advantage of being a little more foolproof in terms of the moment anyone clones the repo or checks out a branch, they _already_ have the exact desired dependencies and are ready to build or test with no further adieu.

Dockerfile
```diff
@@ -1,7 +1,13 @@
-FROM alpine:3.8
+FROM quay.io/deis/lightweight-docker-go:v0.7.0
```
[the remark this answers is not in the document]
krancour: The name of the image is misleading. (Can't recall the genesis of the name now.) This is just Go bundled with some common Go-related tools. It's been used broadly across many projects and source is available at github.com/deis/lightweight-docker-go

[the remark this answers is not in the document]
krancour: Also, you could use a different Go image here. I opted to use the same one that is used for _other_ tasks like linting and unit tests so that we can minimize the set of images we're dependent on and so we can also have a single image that we can point to as being "our development environment."

glyn:
> > I'd like a quick/memorable way to make bin/duffle on macOS or linux.
> 
> See doc changes. There is a way, but perhaps not adequately quick or memorable. I would be happy to add a few OS-specific convenience targets.

That would be really helpful. I'd like something like `make duffle` or `make darwin` to produce `bin/duffle` (without any os/arch naming).

Dockerfile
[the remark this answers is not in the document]
silvin-lubecki: >This is just Go bundled with some common Go-related tools.

Ok great 👍

>  I opted to use the same one that is used for other tasks like linting and unit tests so that we can minimize the set of images we're dependent on 

Makes sense 👍

Makefile
```diff
@@ -1,101 +1,127 @@
...
+################################################################################
+# Build / Publish                                                              #
+################################################################################
+
+.PHONY: build
+build: build-all-bins build-image
+
+.PHONY: build-all-bins
+build-all-bins:
+	$(GO_DOCKER_CMD) bash -c "LDFLAGS=\"$(LDFLAGS)\" scripts/build.sh"
```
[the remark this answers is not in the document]
glyn: > Yep... there a mistake on this line. The prefix here should just be `DOCKER_CMD`. `GO_DOCKER_CMD` is some copy/pasta and is always undefined and @glyn, this is why you were still needing Go on your path. I will fix this as soon as I get to a proper keyboard.

Thanks. But actually I like being able to do a "local" (i.e. non-docker) build using my Go tooling, so please can we preserve that, at least when just building `bin/duffle` for the current OS?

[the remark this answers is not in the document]
krancour: You can opt out of it with `SKIP_DOCKER=true`.

docs/developing.md
krancour: Ok... more thought into this...

Rather than include a target that won't work across all OSes, can I suggest contributors who previously took advantage of this target may actually get _more_ mileage from a symlink from `/usr/local/bin/duffle` (or elsewhere on the path) to `$GOPATH/src/github.com/deislabs/duffle/bin/duffle-<OS>-amd64`?

With this approach, there is no separate step required to install after rebuilding.

krancour:
> That would be really helpful. I'd like something like make duffle or make darwin to produce bin/duffle (without any os/arch naming).

@glyn I added a `build-%` target. You can build for your OS of choice with `make build-<OS>`. It assumes `amd64` as the architecture.

> without any os/arch naming

That makes me uneasy. The same build script that produces binaries on-demand for you also produces deployable artifacts in our pipeline. I think it's best if those artifacts are named unambiguously. This being said...

Elsewhere we discussed `make install` going away and I suggested creating a symlink instead. If you do so, there's no inconvenience to binaries having OS and architecture embedded in their name.

Does that seem ok?

glyn:
> > That would be really helpful. I'd like something like make duffle or make darwin to produce bin/duffle (without any os/arch naming).
> 
> @glyn I added a `build-%` target. You can build for your OS of choice with `make build-<OS>`. It assumes `amd64` as the architecture.
> 
> > without any os/arch naming
> 
> That makes me uneasy. The same build script that produces binaries on-demand for you also produces deployable artifacts in our pipeline. I think it's best if those artifacts are named unambiguously. This being said...
> 
> Elsewhere we discussed `make install` going away and I suggested creating a symlink instead. If you do so, there's no inconvenience to binaries having OS and architecture embedded in their name.
> 
> Does that seem ok?

Yes. I can use a symlink `duffle -> duffle-darwin-amd64`.

docs/developing.md
glyn: Yes, makes sense.

glyn:
I re-tested and `SKIP_DOCKER=true make build-darwin` works as expected as does `make build-darwin`. I wonder if you would mind documenting this in the developer's guide, for posterity?

krancour:
> Looks great now. Ideally, I'd like make duffle-darwin et al. described in the developer's guide, for posterity.

Yes. I will add that.

brigade.js
```diff
@@ -79,161 +211,52 @@ class Notification {
...
-      CHECK_DETAILS_URL: this.detailsURL,
-      CHECK_EXTERNAL_ID: this.externalID
-    }
-    return j.run();
+    this.count++;
+    // Here we are using the mutable 'edge' version of this utility
+    // as an exercise of vetting the current master version of the code in this repo.
+    // It is recommended that immutable tags be used in other cases,
+    // e.g., a proper semver tag of 'vX.X.X' or the short git sha of a particular commit.
+    var job = new Job(`${ this.name }-notification-${ this.count }`, "brigadecore/brigade-github-check-run:edge");
```
vdice: Fwiw, I've been in the habit of specifying `latest` (latest tagged release) for all non-`brigadecore/brigade-github-app` repos (in other words, only really using edge for the repo that produces the artifact).  Another option, of course, is the most recent immutable semver tag, `v0.1.0`.

krancour: Nice catch @vdice. Clearly, you can tell exactly _where_ that was copied/pasted from. Fixing this!

krancour: Fixed.

radu-matei:
Just tested on Windows:

```
PS C:\projects\go\src\github.com\deislabs\duffle> $env:SKIP_DOCKER
true
PS C:\projects\go\src\github.com\deislabs\duffle> make -f .\Makefile build
bash -c "LDFLAGS=\"-w -s -X github.com/deislabs/duffle/pkg/version.Version=de5062b\" scripts/build.sh"
building Windows_NT-amd64
cmd/go: unsupported GOOS/GOARCH pair Windows_NT/amd64
make: *** [build-all-bins] Error 2
```

Of course, this assumes WSL is configured on the machine, which is not default.

krancour:
@radu lemme look at that. This is _really_ odd. I'm not sure how `Windows_NT` got set as.the OS here. Do you have an existing env var by that name?

krancour:
Ok. Apparently the `OS` environment variable (which I am using) is pre-defined as `Windows_NT` on Windows (but not in WSL). So, if not executing within a container, in a place where I expected `OS` to either be a valid OS for Go cross-compiling _or_ be undefined, it is neither of those things. I can fix this.

krancour:
@radu-matei the latest issue you reported should be fixed by the latest commit.

technosophos:
This currently has 4 LGTMs by core maintainers, so I am merging.

PR #776, opened 2019-06-17
ci(brigade.js): run test GH Checks for master branch

vdice:
This removes the `if not master` check in `runSuite` so that Brigade runs the same test checks on master (and, thus, updates this branch's status at https://github.com/deislabs/duffle/branches properly).


vdice:
@glyn agreed.  This is mostly cosmetic.

PR #777, opened 2019-06-18
Use standard Go 1.12 image for docker builds

glyn:
This allows Go 1.12 features to be used and so we revert an earlier commit.

An alternative would be to bump `quay.io/deis/lightweight-docker-go` to use a later version of Go, but it's probably safer to get out of that game and use the more widely supported golang image instead. Also it's not obvious how to make `quay.io/deis/lightweight-docker-go` usable by all maintainers.

Note that `SKIP_DOCKER=true` is currently undocumented and so there is no action for users to take. 
(Those who use `SKIP_DOCKER=true` should know what they are doing.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/deis/lightweight-docker-go:v0.7.0
+FROM golang:1.12-alpine
 ARG LDFLAGS
 ENV CGO_ENABLED=0
 WORKDIR /go/src/github.com/deislabs/duffle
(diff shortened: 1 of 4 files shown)


carolynvs:
Something to keep in mind as we pick Go versions is that this will impact downstream consumers of Duffle (and the same will apply to cnab-go). Porter is on Go 1.11, so moving Duffle to Go 1.12 will force Porter (and anyone else who is relying on Duffle) to move as well, and then that cascades to their consumers. Not a huge impact now, but over time I hope there will be more people using these tools. 😀 

We should articulate how we pick versions and when we will move to newer versions, since both Duffle and cnab-go are the reference implementations that people in the cnab ecosystem rely upon.

To be clear, I am all for switching to a another base image, I am only asking that we put some planning into the Go version bump.

glyn:
Deferring for now.

Issue #772, opened 2019-06-17
Support for relocation mapping files


glyn:
The CNAB spec is being [changed](https://github.com/deislabs/cnab-spec/issues/195) so that relocating a CNAB will no longer produce a new CNAB but will produce a relocation mapping file instead. This file will then be mounted when actions are performed so that the invocation image can determine the original and new references of each image.

Given that [`bundle.json` will also be mounted](https://github.com/deislabs/cnab-spec/issues/144), the relocation mapping file need only contain a mapping from original image reference to new image reference for each image referenced by the CNAB.

PR #778, opened 2019-06-18
Support for relocation mapping files

glyn:
Fixes https://github.com/deislabs/duffle/issues/772

### DIFF
--- a/cmd/duffle/main.go
+++ b/cmd/duffle/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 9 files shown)


cmd/duffle/main.go
```diff
@@ -111,7 +112,46 @@ func prepareDriver(driverName string) (driver.Driver, error) {
...
+
+type driverWithRelocationMapping struct {
+	driver     driver.Driver
+	relMapping string
+}
+
+func (d *driverWithRelocationMapping) Run(op *driver.Operation) error {
+	// if there is a relocation mapping, ensure it is mounted
+	if d.relMapping != "" {
+		fmt.Printf("mounting relocation mapping %q\n", d.relMapping)
```
technosophos: I believe this should be `fmt.Fprintf(os.Stdout...` or it should be removed.

cmd/duffle/relocate.go
```diff
@@ -106,18 +107,14 @@ func (r *relocateCmd) run() error {
 		return err
 	}
 
-	return r.writeBundle(bun)
+	return nil
```
technosophos: This whole section (lines 106-110) should be abbreviated to `return r.relocate(bun)`.

cmd/duffle/main.go
glyn: Oops, I forgot to delete that. Thanks for spotting this.

cmd/duffle/relocate.go
glyn: Yup - another left-over I forgot to clean up. The linter didn't spot it.

PR #782, opened 2019-06-19
thick bundle relocation with relocation mapping files

glyn:


cmd/duffle/export.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/deislabs/duffle/pkg/imagestore/builder"
+
```
silvin-lubecki: nit: empty line

cmd/duffle/relocate.go
```diff
@@ -98,92 +106,78 @@ duffle relocate path/to/bundle.json --relocation-mapping path/to/relmap.json --r
 }
 
 func (r *relocateCmd) run() error {
-	bun, err := r.setup()
+	relMap := make(map[string]string)
+
+	rel, _, cleanup, err := r.setup()
```
silvin-lubecki: It looks like it's the only call to `r.setup()`, and the second argument is dropped, so I guess it can safely been removed from the returned arguments?

cmd/duffle/export.go
glyn: Fixed.

cmd/duffle/relocate.go
glyn: Yes, well spotted. That was a vestige of the earlier form of relocation which produced a modified bundle. Fixed.

Gopkg.toml
```diff
@@ -57,7 +57,7 @@
 
 [[constraint]]
   name = "github.com/pivotal/image-relocation"
-  revision = "532dd0b42e7a50010d7868364309cd314a2bb376"
+  revision = "3c9c32bb3d97fc213476f1c9846a99ca35ecba5d"
```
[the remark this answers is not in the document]
glyn: Updated the dependency to v0.1 (actually the same revision) which includes all changes except one recent doc change.

cmd/duffle/relocate_test.go
```diff
@@ -20,28 +24,43 @@ const (
...
+			t.Fatalf("archiveDir was %q, expected it to end with %q", archiveDir, expectedSuffix)
+		}
+	})
 }
 
-func TestRelocateFileToFileMutatingDigests(t *testing.T) {
-	relocateFileToFile(t, false, fmt.Errorf("digest of image %s not preserved: old digest %s; new digest %s", originalImageNameB, originalImageDigestB, relocatedImageDigestB))
+func TestRelocateFileToFileUnsupportedImageType(t *testing.T) {
+	relocateFileToFile(t, "testdata/relocate/bundle-with-unsupported-image-type.json", errors.New("cannot relocate image c with imageType c: only oci and docker image types are currently supported"), func(archiveDir string) {
+		if archiveDir != "" {
```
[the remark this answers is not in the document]
glyn: No, unless the runtime code fails and creates an archive directory when it shouldn't. The point of this `if` statement is to fail the test in that case.

pkg/imagestore/ocilayout/ocilayout.go
```diff
@@ -0,0 +1,109 @@
...
+	return &Builder{
+		registryClient: registry.NewRegistryClient(),
+		archiveDir:     "",
+		logs:           ioutil.Discard,
+	}
+}
+
+func LocateOciLayout(archiveDir string) (imagestore.Store, error) {
+	layoutDir := filepath.Join(archiveDir, "artifacts", "layout")
+	if _, err := os.Stat(layoutDir); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
glyn: Yes, and that's deliberate. If the `artifacts/layout` directory exists, which is expected to be the case when `LocateOciLayout` is called, then the method will proceed to use that directory and if there are errors (such as a permission error) accessing it, then an error will be returned at that point.

pkg/imagestore/builder/builder.go
```diff
@@ -0,0 +1,69 @@
...
+		logs:       ioutil.Discard,
+	}
+}
+
+type locatingBuilder struct {
+	archiveDir string
+	logs       io.Writer
+}
+
+func (b *locatingBuilder) ArchiveDir(archiveDir string) imagestore.Builder {
```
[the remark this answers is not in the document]
glyn: Because it's implementing the builder pattern. The builder enables an image store to be created with optional archive directory and log stream.

pkg/imagestore/remote/remote.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (b *Builder) Build() (imagestore.Store, error) {
+	return &remote{
+		registryClient: b.registryClient,
+	}, nil
+}
+
+func (r *remote) Add(im string) (string, error) {
+	return "", nil
```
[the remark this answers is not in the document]
glyn: Yes, the `Add` method is called during `duffle export ... --thin` and it mustn't return an error otherwise exporting will fail.

The point of the image store abstraction is to allow both thin bundles (with an image store that keeps images in remote repositories) and thick bundles (with an image store that keeps images in an OCI image layout) to be handled by common code in exporting and relocation operations.

- `Add` is used during exporting. For a thick bundle, it stores an image in the thick bundle's OCI image layout. For a thin bundle, it doesn't need to do anything since the image is already available in a remote repository.
- The other method of the image store interface, `Push`, is used during relocation. For a thick bundle, it copies an image from the thick bundle's OCI image layout to the relocated repository. For a thin bundle, it copies an image from its original repository to the relocated repository.

(However, thanks for asking this question: I deleted some dead code as a result.)

pkg/imagestore/builder/builder.go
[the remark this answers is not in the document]
glyn: @carolynvs @silvin-lubecki also queried this, so I switched to another approach. See the latest commit.

silvin-lubecki:
@glyn please rebase

cmd/duffle/relocate.go
```diff
@@ -98,92 +106,78 @@ duffle relocate path/to/bundle.json --relocation-mapping path/to/relmap.json --r
...
+	dest := ""
+	bundleFile, err := resolveBundleFilePath(r.inputBundle, r.home.String(), r.bundleIsFile)
+	if err != nil {
+		return nil, nop, err
+	}
+
+	var bun *bundle.Bundle
+
+	if strings.HasSuffix(bundleFile, ".tgz") {
+		source, err := filepath.Abs(bundleFile)
```
silvin-lubecki: nit: I guess it would be more readable to extract this whole block its own sub function, like `func extractZippedBundle` ?
And what about changing this `if` with a switch case?
```go
switch{
case strings.HasSuffix(bundleFile, ".tgz"):
    dest, bun, err = extractZippedBundle()
    if err != nil{
         return nil, nop, err
    }
default:
    bun, err = loadBundle(bundleFile)
    if err != nil{
         return nil, nop, err
    }
}
```

cmd/duffle/relocate_test.go
```diff
@@ -2,16 +2,19 @@ package main
...
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
 	"reflect"
+	"strings"
 	"testing"
 
+	"github.com/deislabs/duffle/pkg/imagestore"
+	"github.com/deislabs/duffle/pkg/imagestore/imagestoremocks"
```
silvin-lubecki: nit: the go import grouping is weird, there's duffle/home package alone line 19

pkg/imagestore/construction/construction.go
```diff
@@ -0,0 +1,45 @@
...
+// NewLocatingConstructor creates an image store constructor which will, if necessary, find existing archive contents.
+func NewLocatingConstructor() imagestore.Constructor {
+	return func(options ...imagestore.Option) (imagestore.Store, error) {
+		parms := imagestore.Create(options...)
+		if thin(parms.ArchiveDir) {
+			return remote.Create()
+
+		}
+		return ocilayout.LocateOciLayout(parms.ArchiveDir)
+
```
silvin-lubecki: nit: empty line

pkg/imagestore/construction/construction.go
```diff
@@ -0,0 +1,45 @@
...
+	return ocilayout.Create, nil
+}
+
+// NewLocatingConstructor creates an image store constructor which will, if necessary, find existing archive contents.
+func NewLocatingConstructor() imagestore.Constructor {
+	return func(options ...imagestore.Option) (imagestore.Store, error) {
+		parms := imagestore.Create(options...)
+		if thin(parms.ArchiveDir) {
+			return remote.Create()
+
```
silvin-lubecki: nit: empty line

cmd/duffle/relocate_test.go
glyn: Fixed.

cmd/duffle/relocate.go
glyn: Extracted a function - good idea. I'll defer using a switch statement until there are other cases to consider.

pkg/imagestore/construction/construction.go
glyn: Fixed.

pkg/imagestore/construction/construction.go
glyn: Fixed.

Issue #785, opened 2019-06-20
duffle install should apply any relocation mapping to the invocation image reference


glyn:
https://github.com/deislabs/duffle/issues/772 didn't cover this, but ideally duffle will run the relocated invocation image rather than attempt to run the original (which may not be accessible, e.g. in an air gapped environment).

PR #787, opened 2019-06-21
Apply relocation mapping to invocation image

glyn:
Fixes https://github.com/deislabs/duffle/issues/785


glyn:
Unstacking as https://github.com/deislabs/duffle/pull/798

PR #788, opened 2019-06-24
Bump cnab-go dependency to reflect the Bundle structure changes

jcsirot:
- split parameters into parameters and definitions
- added JSON Schema support for parameter validation
- replaced bundle.Location by the dedicated struct bundle.Credential


jeremyrickard:
Hey @jcsirot, thanks for opening this. I was in the process of doing this, so you're saving the day for me! The only thing I think we should change here is adding a dependency on a released version of cnab-go. I just did a release to include the nil pointer de-reference change that you submitted, so ideally we should add a constraint on that verison of cnab-go here, instead of keeping up with master..something like

```

jcsirot:
@jeremyrickard Indeed, good point. I fixed that.

PR #794, opened 2019-07-10
 keep bundle and manifest in sync

### LINKED ISSUE #793: parameters not recognized
create a bundle:
```
duffle create foo
```
add definition and parameter section to duffle.json:
```
"definitions": {
    "manifest_file": {
      "default": "/cnab/app/kab/manifest.yaml",
      "type": "string"
    }
  },
  "parameters": {
    "fields": {
      "param1": {
        "definition": "manifest_file",
        "destination": {
          "env:": "MANIFEST_FILE"
        }
      }
    }
  }
```
build the bundle:
```
duffle build .
```
specifying the parameter results in the following error:
```
duffle install myfoo foo -s param1=/some/path.foo
Error: parameter param1 not defined in bundle
```
This is despite param1 being present in the bundle definition:
```
cat ~/.duffle/bundles/9460e8557e16656f0f35b991eeb376e1a4086283 | jq
...
  "parameters": {
    "fields": {
      "param1": {
        "definition": "manifest_file",
        "destination": {}
      }
    }
  },
```

sbawaska:
copy all fields from manifest to bundle to ensure that bundle is not
missing any configuration that the bundle author intended.
Added unit tests to keep the two structure in sync.

closes #793

### DIFF
--- a/pkg/builder/builder_test.go
+++ b/pkg/builder/builder_test.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 
 	"github.com/deislabs/cnab-go/bundle"
-
+	"github.com/deislabs/cnab-go/bundle/definition"
 	"github.com/deislabs/duffle/pkg/duffle/manifest"
 	"github.com/deislabs/duffle/pkg/imagebuilder"
 )
@@ -51,18 +51,26 @@ func (tc testImage) Build(ctx context.Context, log io.WriteCloser) error {
 }
 
 func TestPrepareBuild(t *testing.T) {
+	outputs := &bundle.OutputsDefinition{
+		Fields: map[string]bundle.OutputDefinition{"output1": {}},
+	}
+	params := &bundle.ParametersDefinition{
+		Fields: map[string]bundle.ParameterDefinition{"param1": {}},
+	}
 	mfst := &manifest.Manifest{
-		Name:          "foo",
-		Version:       "0.1.0",
-		SchemaVersion: "v1.0.0",
+		Actions: map[string]bundle.Action{"act1": {}},
+		Credentials: map[string]bundle.Credential{"cred1": {}},
+		Custom: map[string]interface{}{"cus1": nil},
+		Definitions: map[string]*definition.Schema{"def1": {}},
 		Description:   "description",
-		Keywords:      []string{"test"},
+		Images: map[string]bundle.Image{"img1": {}},
 		InvocationImages: map[string]*manifest.InvocationImage{
 			"cnab": {
 				Name:          "cnab",
 				Configuration: map[string]string{"registry": "registry"},
 			},
 		},
+		Keywords:      []string{"test"},
 		Maintainers: []bundle.Maintainer{
 			{
 				Name:  "test",
@@ -70,6 +78,11 @@ func TestPrepareBuild(t *testing.T) {
 				URL:   "https://test.com",
 			},
 		},
+		Name:          "foo",
+		Outputs: outputs,
+		Parameters: params,
+		SchemaVersion: "v1.0.0",
+		Version:       "0.1.0",
 	}
 
 	components := []imagebuilder.ImageBuilder{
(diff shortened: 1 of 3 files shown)


pkg/builder/builder_test.go
```diff
@@ -86,21 +100,107 @@ func TestPrepareBuild(t *testing.T) {
...
+		t.Errorf("expected schemaVersion %v, got %v", mfst.SchemaVersion, b.SchemaVersion)
+	}
+	checksPerformed++
+
+	if b.Version != mfst.Version {
+		t.Errorf("expected version %v, got %v", mfst.Version, b.Version)
+	}
+	checksPerformed++
+
+	mfstFields := getFields(manifest.Manifest{})
```
glyn: This is a reasonable check until such time as the fields deliberately diverge. I think it's worth commenting to that effect.
```suggestion
	// Ensure that all the fields have been checked. If the structures need to diverge in the future, this test should be modified.
	mfstFields := getFields(manifest.Manifest{})
```

pkg/builder/builder_test.go
```diff
@@ -86,21 +100,107 @@ func TestPrepareBuild(t *testing.T) {
...
+	if len(mfstFields) != checksPerformed {
+		t.Errorf("expected to check %v fields for equality, but checked only %v fields", len(mfstFields), checksPerformed)
+	}
+}
+
+func TestBundleAndManifestHasSameFields(t *testing.T) {
```
glyn: I'm not convinced that it's worth checking that the field names of the structures are the same since the field names do not actually have to be the same and it seems a bit odd comparing the names of fields with distinct types. Given the previous test, I think we could safely delete this one.

glyn: nit: if we decide to keep this test, the name should be changed:
```suggestion
func TestBundleAndManifestHaveSameFields(t *testing.T) {
```

pkg/builder/builder_test.go
```diff
@@ -86,21 +100,107 @@ func TestPrepareBuild(t *testing.T) {
...
+
+	b := bundle.Bundle{}
+	bundleFields := getFields(b)
+
+	if !reflect.DeepEqual(bundleFields, mfstFields) {
+		t.Errorf("manifest and bundle have different fields.\nmanifest: %+v\nbundle: %+v\n", mfstFields, bundleFields)
+	}
+}
+
+func getFields(i interface{}) map[string]bool {
```
glyn: It's conventional to use `struct{}` as a value type when representing sets as maps in Go.
```suggestion
func getFields(i interface{}) map[string]struct{} {
```

pkg/builder/builder_test.go
```diff
@@ -86,21 +100,107 @@ func TestPrepareBuild(t *testing.T) {
...
+	fields := make(map[string]bool, 15)
```
glyn:
```suggestion
	fields := make(map[string]struct{}, 15)
```

pkg/builder/builder_test.go
```diff
@@ -86,21 +100,107 @@ func TestPrepareBuild(t *testing.T) {
...
+
+	v := reflect.ValueOf(i)
+	typeOf := reflect.TypeOf(i)
+
+	for i := 0; i < v.NumField(); i++ {
+		fields[typeOf.Field(i).Name] = true
```
glyn:
```suggestion
		fields[typeOf.Field(i).Name] = struct{}{}
```

pkg/builder/builder_test.go
sbawaska: > I'm not convinced that it's worth checking that the field names of the structures are the same since the field names do not actually have to be the same and it seems a bit odd comparing the names of fields with distinct types. Given the previous test, I think we could safely delete this one.

In my previous PR: https://github.com/deislabs/duffle/pull/790 the SchemaVersion field was missing in the manifest, but yet elsewhere in duffle a check was performed to see if this field existed and was set. Hence this test. 

IMHO having both bundle and manifest is a bit redundant, there really should be only one struct, but as mentioned in the slack: https://cloud-native.slack.com/archives/CEX1W7WMD/p1562696449224400 InvocationImage is the only field that is different.

glyn: I'm comfortable working on the assumption that the field names should be the same in both structures until such time as we need to change that. It _should_ be obvious that the test is overly restrictive.

PR #798, opened 2019-07-15
Apply relocation mapping to invocation image

glyn:
Fixes https://github.com/deislabs/duffle/issues/785

### DIFF
--- a/cmd/duffle/main.go
+++ b/cmd/duffle/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


cmd/duffle/main.go
```diff
@@ -130,17 +131,38 @@ type driverWithRelocationMapping struct {
...
 
 func (d *driverWithRelocationMapping) Handles(it string) bool {
 	return d.driver.Handles(it)
 }
 
+func (d *driverWithRelocationMapping) relocateImage(im *string) error {
+	relMap := make(map[string]string)
+	err := json.Unmarshal([]byte(d.relMapping), &relMap)
+	if err != nil {
+		return err
```
carolynvs: I suggest wrapping this error to provide some context on where the error came from. Otherwise if it fails on a json unmarshal, the user may get a pretty unhelpful error mesage and it won't be clear that it is due to the relocation mapping.

glyn: Good idea. Fixed.

cmd/duffle/main.go
```diff
@@ -130,17 +131,38 @@ type driverWithRelocationMapping struct {
...
 	}
+
 	return d.driver.Run(op)
 }
 
 func (d *driverWithRelocationMapping) Handles(it string) bool {
 	return d.driver.Handles(it)
 }
 
+func (d *driverWithRelocationMapping) relocateImage(im *string) error {
```
silvin-lubecki: I guess it's just a matter of taste, but what do you think if we could avoid passing mutable parameters, and return the new relocated image instead?
```go
func (d *driverWithRelocationMapping) relocateImage(im string) (string, error) {
```

glyn: Happy to oblige. Pushed a fix.

Issue #811, opened 2019-07-25
Update relocation guide


glyn:
Now that thick bundle relocation is merged, the guide needs updating.

Issue #233, opened 2018-09-25
validate digest on duffle install


michelleN:

glyn:
This needs to take into account any relocation mapping provided to the install.

jeremyrickard:
👍

jeremyrickard:
This specifically needs to handle the relocation aspect. Since we have moved drivers into cnab-go, the business logic for the validation needs to be handled by https://github.com/deislabs/cnab-go/issues/101

PR #812, opened 2019-07-25
Update relocation guide for thick bundles

glyn:
Fixes https://github.com/deislabs/duffle/issues/811


trishankatdatadog:
Thousands of years from now, archeologists will be decoding CNAB bundles...

docs/guides/relocation-guide.md
```diff
@@ -80,8 +80,31 @@ duffle install forge.json --bundle-is-file --relocation-mapping relmap.json ...
...
+```bash
+duffle relocate sort.tgz --bundle-is-file --repository-prefix=registry.gold.gwb.dia/griphook --relocation-mapping relmap.json
+
+```
+This loads the images from `sort.tgz` into the private registry. Relocating from a thick bundle does not need
+access to the original image repositories (which would prevent it from running inside the GWB data center).  
+
+They can now install the sorting software using the original bundle together with the relocation mapping file:
+```bash
+duffle install sort.tgz --bundle-is-file --relocation-mapping relmap.json ...
```
trishankatdatadog: There are some interesting questions this PR should answer, such as:

1. If a `bundle.json` (which contains the original repository names) is signed, how should be a bundle runtime change its behavior given `relmap.json`? Should it ignore the original repository names even though they were signed as part of `bundle.json`?
2. What should a bundle runtime do if the original and new digests do _not_match?

I'm sure there are more questions, but these come to mind right now.

Cc @SantiagoTorres

radu-matei: Context for 1) in https://github.com/deislabs/cnab-spec/issues/195 and https://github.com/deislabs/cnab-spec/pull/218.

For 2), if the content digest differs, then runtimes must stop the execution - I'd have to check to make sure this is explicitly stated in the specification.

I'll also let @glyn comment on this.

radu-matei: To track checking the digest on install - https://github.com/deislabs/duffle/issues/233

docs/guides/relocation-guide.md
```diff
@@ -80,8 +80,31 @@ duffle install forge.json --bundle-is-file --relocation-mapping relmap.json ...
...
+Gringotts Wizarding Bank (GWB) needs to install some software into a new coin sorting machine.
+For GWB, security is paramount. Like Acme, all their production software must be loaded from internal repositories.
+However, GWB regard a networked DMZ as too insecure. Their data center has no connection to the external internet.
+
+Software is delivered to GWB encoded in Base64 and etched on large stones which are then rolled by hand into the
+GWB data center, scanned, and decoded. The stones are stored for future security audits.
+
+GWB obtains the new software as a thick bundle (`sort.tgz`) and relocates it to their private registry as follows:
+```bash
+duffle relocate sort.tgz --bundle-is-file --repository-prefix=registry.gold.gwb.dia/griphook --relocation-mapping relmap.json
```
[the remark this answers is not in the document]
glyn: Ah, `--repository-prefix` determines the private registry that the bundle's images should be relocated to, but `--relocation-mapping` is the path of a file which is _output_ by the relocate operation.

Since this was confusing, I added a paragraph to explain it in the  "duffle relocate" section earlier in the guide.

docs/guides/relocation-guide.md
glyn: What @radu-matei said.

Please note that this PR wasn't really focussed on security and I expect the relocation guide will need to be revisited once the CNAB security spec gets underway.

trishankatdatadog: OK, I'm just worried about any backwards-incompatible changes we might need to make later, and so I'm trying to eliminate them now if we can.

glyn: > OK, I'm just worried about any backwards-incompatible changes we might need to make later, and so I'm trying to eliminate them now if we can.

Understood. I think https://github.com/deislabs/duffle/issues/233#issuecomment-515499707 is about as much as we can do for now, without knowing ahead of time how the security spec is going to turn out.

(Since I had similar concerns, I wrote [Distribution, relocation, content addressing, and signing considerations for CNAB core](https://gist.github.com/glyn/f1e4664747533594503b1404d1d8a9b6) a few weeks ago and discussed it in the CNAB community call.)

Issue #747, opened 2019-05-10
Implement output handling 


michelleN:
https://github.com/deislabs/cnab-spec/pull/133

astrieanna:
It seems like there's a couple things to do here:

1. Store the outputs in a claim (and modify the definition of a claim to make room for that).
2. Add a command to dump the outputs to the screen (`duffle outputs <name of installation>`).

cc @jeremyrickard @ryanmoran @carolynvs @silvin-lubecki

astrieanna:
In order to implement outputs in Duffle, they need to be produced by cnab-go. The tricky part of this is to get the output files from the drivers. In the Kubernetes driver, the Invocation Image is run as a Job. We've come up with a few options for how the driver might get the output files from `/cnab/app/outputs` in the container back to the (k8s client) driver code.

The simplest option would be to use a persistent volume. The driver could mount the volume at `/cnab/app/outputs` in the invocation image and then retrieve the contents after the image has run. However, we don't think the driver should assume or require that persistent volumes are available on all clusters.

The rest of these options assume that an emptyDir volume will be mounted at `/cnab/app/outputs` in the invocation container and that there will be a second "sidecar" container in the same pod that runs the invocation image to do some extra work in retrieving the output files. The containers need to be in the same pod so that they can share the emptyDir volume.

1.  The sidecar container is given a service account token with permission to write secrets. It uses this token to write a secret containing the paths and contents of all files written by the invocation image (when that container finishes running).
  **Challenges**: This requires setting up a service account with the desired permissions; it's not clear that the Duffle driver expects to have those permissions already or that it would be better to ask users to setup this service account in advance.  The sidecar container will also need to figure out when the other container in the pod has finished, so that it can write the final state of the files to the secret.

2. The sidecar container could be used to just keep the pod running, so that the driver could use the API equivalent of `kubectl exec` to grab the output files from the volume.
  **Challenges**: The invocation image is run in a Job; the driver waits for the Job to finish. For the Job to finish, the Pod needs to exit, which means the containers need to exit. This sidecar would prevent that -- meaning a significant restructuring of how the driver waits for invocation image completion would be required.

3. The sidecar container could use its log to output the files from the output volume. This would require choosing some format that can be written to the stdout stream yet is easily parsed into a number of arbitrary binary files.
  **Challenges**: Picking a good format, complicates log stream logic in driver, seems ugly, could hit log size limit edge cases.

[...]

PR #815, opened 2019-07-25
Add `duffle claim show --output`

astrieanna:
## Part of https://github.com/deislabs/duffle/issues/747
After making the code compile against the cnab-go library with outputs support implemented, this PR adds a `--output` flag to `duffle claim show`. This allows users to dump the contents of output files to stdout.

* The output flag takes the name of an output as stated in the bundle; by showing a claim the user can view the names of all the outputs that are part of the bundle.
* Because the cnab-go library implementation returns all the outputs that were generated, even if errors occurred during execution, users will be able to see what output files the invocation image produced during failed runs.

cc @youreddy @jeremyrickard

### DIFF
--- a/cmd/duffle/claims_show.go
+++ b/cmd/duffle/claims_show.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
+	"fmt"
 	"io"
 
 	"github.com/deislabs/cnab-go/claim"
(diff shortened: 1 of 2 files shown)


cmd/duffle/claims_show.go
```diff
@@ -15,42 +17,66 @@ Display the content of a claim.
...
-			storage := claimStorage()
-			return displayClaim(name, w, storage, onlyBundle)
+			cmdData.Name = args[0]
+			cmdData.Storage = claimStorage()
+			return cmdData.runClaimShow(w)
 		},
 	}
 
-	cmd.Flags().BoolVarP(&onlyBundle, "bundle", "b", false, "only show the bundle from the claim")
+	cmd.Flags().BoolVarP(&cmdData.OnlyBundle, "bundle", "b", false, "only show the bundle from the claim")
```
glyn: Why not make this show the bundle from the claim, but not necessarily only the bundle? Then users would mix `-o` and `-b` if they wanted to?

cmd/duffle/claims_show.go
```diff
@@ -15,42 +17,66 @@ Display the content of a claim.
...
+	cmd.Flags().StringVarP(&cmdData.Output, "output", "o", "", "show the contents of the named output")
 
 	return cmd
 }
 
-func displayClaim(name string, out io.Writer, storage claim.Store, onlyBundle bool) error {
-	c, err := storage.Read(name)
+func (csc claimsShowCmd) runClaimShow(w io.Writer) error {
+	if csc.OnlyBundle && csc.Output != "" {
```
glyn: Sometimes these kinds of flag validations are done in a `PreRunE` function.

cmd/duffle/claims_show.go
```diff
@@ -15,42 +17,66 @@ Display the content of a claim.
...
+		return errors.New("invalid flags: at most one of [--bundle, --output] can be specified")
```
glyn: Is the following more readable?
```suggestion
		return errors.New("invalid flags: at most one of --bundle and --output can be specified")
```

Gopkg.toml
```diff
@@ -61,7 +61,8 @@
 
 [[constraint]]
   name = "github.com/deislabs/cnab-go"
-  version = "v0.2.1-beta1"
+  branch = "pr/duffle-outputs"
```
glyn:
```suggestion
  # FIXME: delete the following two lines and reinstate a version constraint before merging
  branch = "pr/duffle-outputs"
```

cmd/duffle/claims_show.go
youreddy: A reason not to enable `-o` and `-b` to be used together is that it is not obvious what such a command should do.

Some options:
* If you think of each of these as a filter on what output should be produced, then specifying both gives you no output (because they filter for different parts of the claim, there is not intersection).
* Since of `-o` and `-b` has a specific output, using both gives you both (in some order).
* Since `-b` says you want part of the bundle and `-o` selects an output listed in the bundle, both gives the specification of the output from the bundle in the claim (rather than the contents of the file).

We tend to think of the CLI behavior through the first mental model (filtering), so the use-case of combining the flags is very limited. However, if users did want to mix `-o` and `-b`, we should decide what that behavior should be and how to easily convey that to the user.

astrieanna:
@silvin-lubecki does this side of the PR look good?

cmd/duffle/claims_show.go
silvin-lubecki: I was also wondering if `--output` could be a string slice, and then print all the selected outputs ? 🤔

youreddy: I'm fine fine with either a string slice or the way it's implemented in the PR; both implementations fit our use case. I leave it to the repo owners to decide what the cli user experience should be.

youreddy:
The PR in cnab-go was merged, going to update the Gopkg.toml file to point to the updated version!

jeremyrickard:
@youreddy should we release cnab-go or are you planning on pinning to a commit #? I'd prefer a released version, but we can always update to that once we have a few more things in.

youreddy:
@jeremyrickard, I was going to specify master but you're right we should do a released version. And yes to releasing cnab-go!

astrieanna:
@jeremyrickard /brig run ?

jeremyrickard:
@astrieanna sorry for the delayed response. Were you asking me to re-run it, or what "/brig run" does? Happy to explain what it does if you'd like!

jeremyrickard:
Thanks for the persistence on this @astrieanna and @youreddy, much appreciated.

Issue #817, opened 2019-08-02
certain image names in duffle.json produce spurious images in bundle


glyn:
The following `duffle.json`:
```
{
    "name": "imagebug",
    "version": "1",
    "schemaVersion": "v1.0.0-WD",
    "invocationImages": {
        "cnab": {
            "name": "cnab",
            "builder": "docker",
            "configuration": {
                "registry": "imagebug"
            }
        }
    },
    "images": {
        "x.y": {
            "image": "docker.io/cloudfoundry/cnb@sha256:93e2707b756ed9544ecccab6884424e4990e2b3ec08f0435b8249499b51b2354",
            "digest": "sha256:93e2707b756ed9544ecccab6884424e4990e2b3ec08f0435b8249499b51b2354"
        }
    }
}
```
when input to `duffle build`, produces an invalid images section of bundle.json:
```
{
  "description": "",
  "images": {
    "x": {
      "description": "",
      "image": "",
      "imageType": ""
    }
  },
  "invocationImages": [
    {
      "image": "imagebug/imagebug-cnab:32ff5be1120468ffcb5c37226c5348276b9de209",
      "imageType": "docker"
    }
  ],
  "name": "imagebug",
  "schemaVersion": "v1.0.0-WD",
  "version": "1.0.0"
}
```

This seems superficially like https://github.com/deislabs/duffle/issues/658, which was apparently fixed.

## Possible clue
Change `x.y` to `x` in `duffle.json` and the correct bundle is produced:
```
{
  "description": "",
  "images": {
    "x": {
      "description": "",
      "image": "docker.io/cloudfoundry/cnb@sha256:93e2707b756ed9544ecccab6884424e4990e2b3ec08f0435b8249499b51b2354",
      "imageType": ""
    }
  },
  "invocationImages": [
    {
      "image": "imagebug/imagebug-cnab:32ff5be1120468ffcb5c37226c5348276b9de209",
      "imageType": "docker"
    }
  ],
  "name": "imagebug",
  "schemaVersion": "v1.0.0-WD",
  "version": "1.0.0"
}
```

## Steps to reproduce (at least on macOS)
* Grab `duffle` from https://github.com/deislabs/duffle/releases/tag/0.2.0-beta.3 and add it to your path
* Download and unzip [build.zip](https://github.com/deislabs/duffle/files/3461773/build.zip)
* cd into the unpacked archive and into the `build` subdirectory
* issue `duffle build . -o outbundle.json`
[...]

PR #818, opened 2019-08-02
Replace viper with standard JSON decoding

glyn:
Fixes https://github.com/deislabs/duffle/issues/817


carolynvs:
Viper is still being imported in `cmd/duffle/install.go`. Once you remove that import, you should be able to run `dep ensure` and see it disappear from Gopkg.lock.

glyn:
> Viper is still being imported in `cmd/duffle/install.go`. Once you remove that import, you should be able to run `dep ensure` and see it disappear from Gopkg.lock.

Thanks @carolynvs - that does seem to clean up the dependency...

However, `parseValues` in `cmd/duffle/install.go` depends rather deeply on viper and will need more thought to reword as standard JSON decoding. Any insights into the design of that code would be helpful.

carolynvs:
Here's the original implementation of `parseValues` with just the relevant json part. The function  used to do a switch on the file extension (*.json, *.toml, etc) which we don't need anymore:

```
vals := map[string]interface{}{}
data, err := ioutil.ReadFile(file)
if err != nil {
  return vals, err
}
err = json.Unmarshal(data, &vals)
return vals, err
```

glyn:
> Here's the original implementation of `parseValues` with just the relevant json part.

Thanks @carolynvs - I just found that in git history, but thanks for the confirmation this is still the right approach.

PR #832, opened 2019-08-10
ref(*): remove the loader pkg; use cnab-go's

vdice:
Removes the `loader` pkg per https://github.com/deislabs/cnab-go/pull/98

TODO:
 - [x] replace `vendor` mod with appropriate cnab-go bump (didn't go the route of updating Gopkg.toml w/ my source/branch b/c it's based on latest cnab-go, which includes many changes;  see https://github.com/deislabs/duffle/pull/827)


cmd/duffle/export.go
```diff
@@ -6,9 +6,10 @@ import (
 
 	"github.com/deislabs/duffle/pkg/duffle/home"
 	"github.com/deislabs/duffle/pkg/imagestore/construction"
-	"github.com/deislabs/duffle/pkg/loader"
 	"github.com/deislabs/duffle/pkg/packager"
 
+	"github.com/deislabs/cnab-go/bundle/loader"
```
silvin-lubecki: nit: The modified imports are not well grouped, here and in the following files

Issue #828, opened 2019-08-06
duffle build followed by duffle export should work


itowlson:
If you do a `duffle build` followed by a (full-bundle) `duffle export` then you get a baffling error (per #824), the significance of which is "you need to push the invocation image first," even though you may already have the invocation image in your local Docker registry as a result of `duffle build`.

The reason that Duffle always attempts an image pull is so that it can work in environments where there is no local Docker daemon (thanks @glyn for the explanation).  But this makes for a non-smooth build-then-export experience, not made easier by the cryptic error.

Some possible solutions:

1. `duffle build` could always push the invocation image.  @glyn has pointed out that this avoids problems where the user does a `duffle build` followed by `duffle export -t` (manifest only bundle).  At the moment, the exported thin bundle will Work On My Machine TM, but fail when I give it to someone else because I haven't published the image.  Downside is that maybe I don't want to put my invocation image up in public just yet; it also makes it harder to test locally, e.g. using Kind - these problems could be mitigated by a `duffle build --push-image=false` option.

2. `duffle export` could check whether there is a local Docker daemon, and if so check that before resorting to a pull.

Or something else of course!

itowlson:
Another idea: `duffle export -t` could perform the push - this would address the 'exported thin bundle fails on consumers' machines' problem, and makes some kind of sense because `export` implies an intent to distribute/publish more than `build` does.

This would still require `duffle export` without the `-t` to check the local Docker daemon if there was one, as conceptually there's no need to push in the full bundle scenario - your images are included _in_ the export and don't need to be placed in a separate registry.

glyn:
> ... these problems could be mitigated by a duffle build --push-image=false option

Agreed, it's not always appropriate for `duffle build` to push the image.

> duffle export could check whether there is a local Docker daemon, and if so check that before resorting to a pull.

That feels like a clean solution, but it's non-trivial. We would need to extend pivotal/image-relocation to support that (or, if that's not feasible, see if there is a different library for writing the OCI image layout).

> Another idea: `duffle export -t` could perform the push

I think that behaviour would be rather surprising. Also, the user can issue `duffle build -o` to get a thin bundle without using `duffle export -t` and then we are back to the same issue - if they give that thin bundle to someone using a different docker daemon, the other person can fall foul of the missing invocation image.

glyn:
> duffle export could check whether there is a local Docker daemon, and if so check that before resorting to a pull.
> 
[...]

PR #834, opened 2019-08-12
Load images from docker daemon in export and relocate

glyn:
This is implemented by bumping pivotal/image-relocation to pick up the docker
daemon support.

Fixes https://github.com/deislabs/duffle/issues/828


glyn:
I'm wondering if this change is robust given this observation: https://github.com/deislabs/duffle/pull/691#issuecomment-480893767. I need to investigate.

radu-matei:
I need to investigate that PR again, but from what I recall, we can't rely on the digest calculated locally (as different registries can output different digests for the same content).

> During bundle development, it may be ideal to omit the contentDigest field and/or skip validation. Once a bundle is ready to be transmitted as a thick or thin bundle, it must have a contentDigest field.

Based on a recent discussion in #cnab and to the spec, we proposed adding digest to the images on the first push, if any digest is missing.
(Note that this would be the only instance where `push` would mutate the bundle).

What do you think?

glyn:
Supplying any missing digest on first push is a partial solution, but it doesn't cover the case where the bundle is never pushed. With this PR the following scenario becomes possible:
1. Build a bundle (but don't push the invocation image)
2. Export the bundle as a thick bundle (which will read the image from the daemon)
3. Relocate the thick bundle
4. Install the thick bundle with the relocation mapping from the previous step.

In this case, the image's content digest ends up being that which was derived from the daemon and may (I'm still investigating this) therefore be subject to the risk pointed out in the other PR.

Without this PR, `duffle export` will only succeed if the images have already been pushed and it will use the content digest produced by the (initial) registry.

glyn:
It turns out we can't depend on the docker daemon forming the same digest as a repository. It seems to do at the moment, but that's just (bad IMO) luck. Closing.

PR #846, opened 2019-08-16
Basic docs on how to release

glyn:


docs/developing.md
```diff
@@ -107,3 +107,8 @@ $ make dep
 ## Debugging
 
 For instructions on using VS Code to debug the Duffle binary, see [the debugging document](debugging.md).
+
+## Creating a Release
+
+To create a release, simply push a tag and CI will do the rest.
```
itowlson: You need to push a tag _in a specific format_.  When Jeremy was trying to do a release he pushed ‘v0.3.0-beta.1’ and that didn’t trigger the release.  The docs don’t need to state the format to regex level precision, but maybe give examples, and flag not to use a ‘v’ or ‘v.’ prefix (especially as I gather cnab-go does use a ‘v’ prefix.

glyn: Fixed, thanks.

Issue #845, opened 2019-08-15
`duffle upgrade` fails against thick bundle


ashwin-venkatesh:
Hey!! We ran into an issue when trying to update the instance of our bundle. The command we ran for context is:
```
duffle upgrade ${CLUSTER_NAME} -c ./credentials.yaml \
    --set kubernetes_env="${CLUSTER_NAME}" \
    -m $PWD/relocated.json \
    --bundle-file build-service-*.tgz
```

This failed with the following error message: `Error: cannot load bundle: invalid character '\x1f' looking for beginning of value`.

We were able to work around this by importing the bundle and then running upgrade against the `bundle.json` directly. The duffle version we used was `duffle 0.2.0-beta.3`

youreddy:
Ran into this issue as well!

glyn:
I also noticed that `duffle upgrade` wasn't validating the bundle, so I'll fix that at the same time.

PR #847, opened 2019-08-16
Allow duffle upgrade to accept a thick bundle

glyn:
Also validate the bundle being used to upgrade.

Fixes https://github.com/deislabs/duffle/issues/845

### DIFF
--- a/cmd/duffle/upgrade.go
+++ b/cmd/duffle/upgrade.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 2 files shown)


cmd/duffle/upgrade.go
```diff
@@ -90,13 +91,20 @@ func (up *upgradeCmd) run() error {
 
 	// If the user specifies a bundle file, override the existing one.
 	if up.bundleFile != "" {
-		bun, err := loadBundle(up.bundleFile)
+		bun, tempDir, err := inferAndLoadBundle(up.bundleFile)
```
glyn: @benpatt Your work the other week is paying dividends ^^^

PR #849, opened 2019-08-28
fix(brigade.js): prevent token leak if github-release fails

vdice:
We discovered that the utility used to create a GH release and upload binaries _may_ leak the auth token in an error scenario, hence the need to filter the output.

Tested via intentionally setting one of the env vars that this tool uses to determine repo (`GITHUB_REPO`) to an invalid value and we see:

```
 $ k logs release-01dkcs3gpydtxyh9hhjan6jsp4
bash -c "LDFLAGS=\"-w -s -X github.com/deislabs/duffle/pkg/version.Version=0.3.2-vdicetest\" scripts/build.sh"
building linux-amd64
building windows-amd64
building darwin-amd64
error: github returned 404 Not Found
error: expected '200 OK' but received '404 Not Found' (url: https://api.github.com/repos/vdice/testfakerepo/releases?access_token=<REDACTED>&per_page=100)
error: expected '200 OK' but received '404 Not Found' (url: https://api.github.com/repos/vdice/testfakerepo/releases?access_token=<REDACTED>&per_page=100)
error: expected '200 OK' but received '404 Not Found' (url: https://api.github.com/repos/vdice/testfakerepo/releases?access_token=<REDACTED>&per_page=100)
```

Seemingly the `github-release upload ...` invocation is the culprit for the token leak, but I also added a filter to the preceeding `github-release release ...` invocation.

Additionally, I removed the `|| echo "release ${tag} exists"` as it was rendered moot now that we pipe to `sed` and lose the return code, but this case/cause remains evident from the message returned from `github-release`, e.g.,

```
error: github returned 422 Unprocessable Entity (this is probably because the release already exists)
```

### DIFF
--- a/brigade.js
+++ b/brigade.js
@@ -97,6 +97,7 @@ function githubRelease(p, tag) {
     "GITHUB_REPO": parts[1],
     "GITHUB_TOKEN": p.secrets.ghToken,
   };
+  job.shell = "/bin/bash";
   job.tasks = [
     "go get github.com/aktau/github-release",
     `cd ${localPath}`,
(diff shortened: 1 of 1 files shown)


vdice:
Related note: There is a fix in Brigade to properly set `-o pipefail` if the shell is `/bin/bash` so that these jobs will properly exit non-zero (with added pipeline usage): https://github.com/brigadecore/brigade/pull/982

vdice:
@radu-matei can you review this fix when convenient?

vdice:
I've pushed an amended commit (rebase against master; no other changes) to re-trigger CI...

Issue #859, opened 2019-09-16
Improve how relocation mappings are applied


glyn:
The new support in cnab-go for modifying operations should be exploited.

Issue #858, opened 2019-09-16
status command should support relocation mapping


glyn:
Without this, the original invocation image will be used and this is not necessarily available when images have been relocated.

PR #862, opened 2019-09-16
Improve how relocation mappings are applied

glyn:
Also support a relocation mapping flag on the status command.

Fixes https://github.com/deislabs/duffle/issues/858
Fixes https://github.com/deislabs/duffle/issues/859


cmd/duffle/main_test.go
```diff
@@ -162,3 +164,85 @@ func TestFindCreds(t *testing.T) {
...
+			expectedOpRelocatorErrorMessage: "",
+		},
+		"omitted relocation mapping file": {
+			relMapFile:                      "",
+			invocationImage:                 "example.com/original",
+			expectedErrorMessage:            "",
+			expectedInvocationImage:         "example.com/original",
+			expectedOpRelocatorErrorMessage: "",
+		},
+		"non-existent relocation mapping file": {
```
carolynvs: What is if the difference between this test case and "bad relocation mapping file"? They both seem to have the same setup with a non-existent relocation mapping file, but then they each have different `expectedInvocationImage` which I don't quite understand.

glyn: Well spotted! There is no significant difference, so I've deleted one of them. `expectedInvocationImage` is unused since `expectedErrorMessage` is non-empty and so its value makes no difference.

glyn: Pushed a fix and also improved the test to check the mounting of the relocation mapping.

PR #863, opened 2019-09-16
Add Urvashi Reddy as a maintainer

jeremyrickard:
Pivotal's contributions for outputs in cnab-go and the Duffle driver were very important and they have indicated they will continue to be involved in the spec, as well as Duffle and cnab-go development, so I'd like to nominate Urvashi as a maintainer. She is also a cnab-spec approver!


jeremyrickard:
We've reached a majority, so welcome @youreddy!

PR #864, opened 2019-09-16
Removing Matt Fisher from the CODEOWNERS file

jeremyrickard:
Matt Fisher is no longer actively working on CNAB/Duffle, we thank you for your service @bacongobbler .


jeremyrickard:
> So long @bacongobbler! 👋

It's not farewell, it's see you later.

Issue #867, opened 2019-09-19
duffle relocate should preserve tag when writing to a tagged and digested image reference


glyn:
See https://github.com/pivotal/image-relocation/issues/33 for details. The fix will be to bump image-relocation in duffle once https://github.com/pivotal/image-relocation/issues/33 has been implemented.

PR #869, opened 2019-09-23
duffle relocate preserves tag of a tagged and digested image reference

glyn:
... when writing to such a reference

Fixes https://github.com/deislabs/duffle/issues/867

### DIFF
--- a/vendor/github.com/pivotal/image-relocation/pkg/registry/ggcr/remote.go
+++ b/vendor/github.com/pivotal/image-relocation/pkg/registry/ggcr/remote.go
@@ -86,7 +86,7 @@ func writeRemoteImage(i v1.Image, n image.Name) error {
 		return err
 	}
 
-	ref, err := name.ParseReference(n.String(), name.WeakValidation)
+	ref, err := getWriteReference(n)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


vendor/github.com/pivotal/image-relocation/pkg/registry/ggcr/remote.go
```diff
@@ -119,3 +119,13 @@ func resolve(n image.Name) (authn.Authenticator, error) {
 
 	return resolveFunc(repo.Registry)
 }
+
+func getWriteReference(n image.Name) (name.Reference, error) {
+	// if target image reference is both tagged and digested, ignore the digest so the tag is preserved
```
silvin-lubecki: I thought that it was the opposite? If the image reference has both `tag` and `digest`, then `tag` is skipped and only the `digest` is used when pulled ? I think that this behavior is due to the fact that a `tag` can be "forced pushed", so the digest may differ ? Maybe am I missing something?

glyn: You are quite right for the behaviour when _pulling_ an image, but this change relates to the target reference when copying an image, which is analogous to _pushing_ an image. Once we've copied an image in this way, it can be referenced using either the tag or the digest.

silvin-lubecki: Ok right, thank you @glyn for the clarification!

Issue #830, opened 2019-08-07
Should duffle run require a claim name on stateless actions?


itowlson:
`duffle run` currently always requires a claim name (installation name).

However, the CNAB spec for the `actions` section (https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#custom-actions) says:

"`stateless`: The action *does not act on a claim*, and does not require credentials. This is useful for exposing dry-run actions, printing documentation, etc." (emphasis added)

Should the claim name be optional in `duffle run` if the action is stateless?  For example, how should the user perform a "dry run" custom action if they have never installed the bundle?

glyn:
Makes sense to me. Note that the spec also says:
> Stateless actions can be invoked on a non-existing installation.

which I think is ambiguous. It could mean that an existing installation need not be supplied or that the installation name can be one that doesn't exist. I think the former makes more sense.

itowlson:
I wonder if the spec is leaving open the possibility of a speculative "print a dry run of what would happen _if_ I supplied this installation name" action.  But I agree it seems confusing, and could make implementations more complicated.

I have raised issue https://github.com/deislabs/cnab-spec/issues/252.

PR #870, opened 2019-09-27
Enable custom actions on existing claims or bundles

youreddy:
After reading [this issue comment](https://github.com/deislabs/duffle/issues/629#issuecomment-462829427) I took a look at how we could make it possible to run custom actions on bundles that duffle has built but not installed or bundles it does not know about. Since running a custom action currently requires a claim, the easiest thing to do was create a claim if one did not exist. 

This PR allows `duffle run` to perform custom actions in the following scenarios

1. When the claim exists: 
users can run `duffle run someCustomAction--claim existingClaim` and duffle will reference the claim store in order to perform the custom action. 

2. When the claim does not exist but duffle knows about the bundle: 
users can run `duffle run someCustomAction --bundle someBundle --claim myNewClaim` and duffle will look in the store to perform the custom action. It will also create a claim if the action is a modifying one. 

3. When the claim doe not exist and duffle does not know about the bundle: 
users can run `duffle run someCustomAction --bundle-is-file some-bundle.json --claim myNewClaim` and duffle will use the bundle definition to run the custom action. It will also create a claim if the action is a modifying one. 

Let me know what you all think of this functionality. I'm open to feedback and restructuring the design of this cli command.  

Related Issues: 
- https://github.com/deislabs/duffle/issues/629
- https://github.com/deislabs/duffle/issues/830

### DIFF
--- a/cmd/duffle/run.go
+++ b/cmd/duffle/run.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"io"
+	"log"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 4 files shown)


youreddy:
Hi @radu-matei @glyn @jeremyrickard @carolynvs! Would be great to get some 👀 on this one when you get a chance.

cmd/duffle/run.go
```diff
@@ -3,107 +3,192 @@ package main
...
 	}
+	run.opOutFunc = setOut(cmd.OutOrStdout())
+
 	flags := cmd.Flags()
-	flags.StringVarP(&driver, "driver", "d", "docker", "Specify a driver name")
-	flags.StringVarP(&relocationMapping, "relocation-mapping", "m", "", "Path of relocation mapping JSON file")
-	flags.StringArrayVarP(&credentialsFiles, "credentials", "c", []string{}, "Specify a set of credentials to use inside the CNAB bundle")
-	flags.StringVarP(&valuesFile, "parameters", "p", "", "Specify file containing parameters. Formats: toml, MORE SOON")
-	flags.StringArrayVarP(&setParams, "set", "s", []string{}, "Set individual parameters as NAME=VALUE pairs")
+	flags.StringVarP(&run.claimName, "claim", "i", "", "Specify the name of an existing claim (required)")
```
glyn: Is this always an existing claim?

cmd/duffle/run.go
```diff
@@ -3,107 +3,192 @@ package main
...
+	flags.StringVarP(&run.bundleName, "bundle", "b", "", "Specify the name of a bundle")
+	flags.StringVarP(&run.bundlePath, "bundle-is-file", "f", "", "Specify the path to a bundle.json")
+	flags.StringVarP(&run.driver, "driver", "d", "docker", "Specify a driver name")
+	flags.StringVarP(&run.relocationMapping, "relocation-mapping", "m", "", "Path of relocation mapping JSON file")
+	flags.StringArrayVarP(&run.credentialsFiles, "credentials", "c", []string{}, "Specify a set of credentials to use inside the CNAB bundle")
+	flags.StringVarP(&run.valuesFile, "parameters", "p", "", "Specify file containing parameters. Formats: toml, MORE SOON")
+	flags.StringArrayVarP(&run.setParams, "set", "s", []string{}, "Set individual parameters as NAME=VALUE pairs")
+
+	err := cmd.MarkFlagRequired("claim")
+	if err != nil {
```
glyn: nit: in duffle, it's more usual to fold assignments and if statements together when no new variables need to last beyond the scope of the if.

cmd/duffle/run.go
```diff
@@ -3,107 +3,192 @@ package main
...
+
+	if r.bundleName != "" {
+		r.bundlePath, err = resolveBundleFilePath(r.bundleName, r.home, false)
+		if err != nil {
+			return err
+		}
+	}
+
+	if r.bundlePath != "" {
+		return r.createClaimFromBundlePath()
```
glyn: What happens here if a claim already exists?

youreddy:
Thanks for taking a look @glyn! 

>The approach of creating new claims just to run custom actions seems to change that and might cause issues.

I was going back and forth on this. Attaching a bundle definition to a new claim was the easiest way to maintain the current interface for creating a driver operation. But I could have also implemented this with a temporary claim that served the purpose of creating the operation and then not persist the claim in duffle's store. This would maintain the constraint that claims should only be created on `install`. I ended up keeping the claim around because it was the only way for a user to keep track of which action had been run and its status. 

The delete implications are valid concerns and I would be happy to make changes that didn't persist the claim in the duffle store. I think that's the right thing to do since the spec specifically defines claims as installed bundles. How would you feel about the PR if I made that change?

>Also, it seems slightly odd that modifying custom actions should save their claims, but non-modifying actions shouldn't as that makes the lifecycle of claims even more complicated to reason about.

This was just to maintain the current logic [here](https://github.com/deislabs/duffle/blob/master/cmd/duffle/run.go#L89-L92).

glyn:
> Thanks for taking a look @glyn!
> 
> > The approach of creating new claims just to run custom actions seems to change that and might cause issues.
> 
> I was going back and forth on this. Attaching a bundle definition to a new claim was the easiest way to maintain the current interface for creating a driver operation.

I've been looking at this and it appears that `runCmd.claimName` will be the empty string in cases where you have to "new up" a claim from the bundle. I think the namespace for claim names is global, so such claims, if persisted, would tend to clobber each other. This doesn't feel good.

> But I could have also implemented this with a temporary claim that served the purpose of creating the operation and then not persist the claim in duffle's store. This would maintain the constraint that claims should only be created on `install`. I ended up keeping the claim around because it was the only way for a user to keep track of which action had been run and its status.

Wouldn't it be better to change the contract and admit that there may not be a claim when a custom action is run against a bundle which is not currently installed?

> 
> The delete implications are valid concerns and I would be happy to make changes that didn't persist the claim in the duffle store. I think that's the right thing to do since the spec specifically defines claims as installed bundles. How would you feel about the PR if I made that change?

I think that would be an improvement, but it may not go far enough (see above).

> 
> > Also, it seems slightly odd that modifying custom actions should save their claims, but non-modifying actions shouldn't as that makes the lifecycle of claims even more complicated to reason about.
> 
> This was just to maintain the current logic [here](https://github.com/deislabs/duffle/blob/master/cmd/duffle/run.go#L89-L92).

Ok. It's not obvious to me that this logic needs to be preserved in this case.

youreddy:
> I've been looking at this and it appears that runCmd.claimName will be the empty string in cases where you have to "new up" a claim from the bundle. I think the namespace for claim names is global, so such claims, if persisted, would tend to clobber each other. This doesn't feel good.

Right, so I think we're moving away from the idea of persisting it and we can use a randomly generated string for the temporary claim name. This would be sufficient for creating the driver operation and would it make it so that the claim name is only required if it already exists. 

>Wouldn't it be better to change the contract and admit that there may not be a claim when a custom action is run against a bundle which is not currently installed?

Can you expand on this? By admit, do you mean providing some log output that the claim does not exist and ask the user to retry with a bundle reference?

glyn:
> > I've been looking at this and it appears that runCmd.claimName will be the empty string in cases where you have to "new up" a claim from the bundle. I think the namespace for claim names is global, so such claims, if persisted, would tend to clobber each other. This doesn't feel good.
> 
> Right, so I think we're moving away from the idea of persisting it and we can use a randomly generated string for the temporary claim name. This would be sufficient for creating the driver operation and would it make it so that the claim name is only required if it already exists.
> 
> > Wouldn't it be better to change the contract and admit that there may not be a claim when a custom action is run against a bundle which is not currently installed?
> 
> Can you expand on this? By admit, do you mean providing some log output that the claim does not exist and ask the user to retry with a bundle reference?

No, I meant that we should admit the facts of the situation to the custom action which will then see an empty string as a claim name and understand that the bundle isn't installed and that there isn't a claim.

Specifically, I think we need to provide a way to create a driver operation when there isn't a claim.

youreddy:
>Specifically, I think we need to provide a way to create a driver operation when there isn't a claim.

Yeah, that was the larger change I was trying to avoid. Especially since test coverage throughout the code base is inconsistent. I'll revisit this PR and see what it would take to make that change.

PR #885, opened 2020-02-06
Run make dep

glyn:
... to fix the fact that make verify-vendored-code was broken at fb912c3c874f4eee362:

  $ SKIP_DOCKER=true make verify-vendored-code
  dep check
  # Gopkg.lock is out of sync:
  github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2017-05-10/resources: imported or required, but
  missing from Gopkg.lock's input-imports
  github.com/Azure/go-autorest/autorest: imported or required, but missing from Gopkg.lock's input-imports
  github.com/Azure/go-autorest/autorest/azure/auth: imported or required, but missing from Gopkg.lock's input-imports


glyn:
CI failed unexpectedly. Closing to avoid people wasting time reviewing.

PR #887, opened 2020-02-14
Update to Go modules and cnab-go v0.8.2-beta1

radu-matei:
depends on #886 

TODO:

- [x] update CI to use Go 1.13
- [x] make sure the updated `brigade.js` works for all targets, not just `test` (cc @vdice)


radu-matei:
Not yet sure what is causing the CI issues: 

```
# github.com/docker/cli/cli/config
/go/pkg/mod/github.com/docker/cli@v0.0.0-20191017083524-a8ff7f821017/cli/config/config.go:109:18: undefined: os.UserHomeDir
Makefile:95: recipe for target 'build-all-bins' failed
make: *** [build-all-bins] Error 2
```

brigade.js
```diff
@@ -1,9 +1,9 @@
 const { events, Job, Group } = require("brigadier");
 
-const projectOrg = "deislabs";
+const projectOrg = "cnabio";
 const projectName = "duffle";
 
-const goImg = "quay.io/deis/lightweight-docker-go:v0.7.0";
+const goImg = "golang:1.13";
```
radu-matei: cc @vdice - I just used the official Go image here - do we want to update original Go image we were using?

carolynvs: +1 for using the official go image

vdice: Official sounds good to me!  I'll test this branch to be sure... can't recall if the official image has `make` for instance...

vdice: (It does have `make` 🎉 )   The other targets look fine; the buildAndPublishImage job worked for my fork off of this branch.  I didn't explicitly test the github release job, but I don't see why that would break.  The `brigade.js` needs an overhaul to slim down and use utilities from the brigade-utils library... we should file a follow-up.

PR #893, opened 2020-03-12
fix(*): docker-based build

vdice:
* Fixes the Docker-based build (transitions to the official `golang` image in the process)
* Updates CI (brigade.js) to add a test/check for this build
* Adds configurability for interactive Docker runs (necessary for non-interactive CI check)
* Cleans up/refactors a bit of brigade.js in the process to utilize updated utilities (good chance to remove custom code)

Question for reviewers:

I wanted to add a check to ensure functionality of the Docker-based build _somewhere_ in CI, to catch failures like this in the future, but am open to only running it on merges to master (and not on every PR, as is currently implemented).  Opinions?

Fixes https://github.com/cnabio/duffle/issues/892


carolynvs:
What I've done in the past is this:

1. Document in the contributing guide what version of go is required/supported for development. This is just the state of affairs unfortunately.
1. Optionally put in a check in the makefile for the go version if you want to get really strict about it and catch people filing issues about using go 1.14 when you only support go 1.13 for example.
1. Run builds on master that check your supported version(s) of go _and_ tip. Tip should never cause the build/release to fail, but you can use it as an early warning to know that you have work ahead of you, and how bad it looks. I recommend putting that into a separate build, maybe run it on a schedule in addition to commits on master.

Makefile
```diff
@@ -22,13 +22,19 @@ BASE_PACKAGE_NAME := github.com/cnabio/duffle
 
 ifneq ($(SKIP_DOCKER),true)
 	PROJECT_ROOT := $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
-	DEV_IMAGE := quay.io/deis/lightweight-docker-go:v0.7.0
-	DOCKER_CMD := docker run \
-		-it \
+	DEV_IMAGE := golang:1.13
```
carolynvs: 🎉 I love seeing us move to supporting base images

brigade.js
```diff
@@ -66,14 +69,36 @@ function test() {
...
 
+function testViaDocker() {
+  // Create a new job to run Go tests via Docker
+  // to ensure Docker-based builds are functional
+  var job = new Job("tests-via-docker", "docker:stable-dind");
+  // Set privileged true to enable Docker-in-Docker
+  job.privileged = true;
+  // Set a few environment variables.
+  job.env = {
+    "GO111MODULE": "on",
```
carolynvs: Is this env var necessary? 🤔  I always set it in the Makefile so that it is set consistently everywhere. But I also see that it's being set in the Dockerfile right now.

brigade.js
```diff
@@ -66,14 +69,36 @@ function test() {
...
+    "DOCKER_INTERACTIVE": "false",
+  };
+  // Run Go unit tests
+  job.tasks = [
+    "apk add --update --no-cache make git > /dev/null",
+    "dockerd-entrypoint.sh > /dev/null 2>&1 &",
+    "sleep 20",
```
carolynvs: What is the magic 20 second sleep waiting for? Can we add a comment for this line please? 😀

brigade.js
vdice: Mmm, appears not.  Removed.

brigade.js
vdice: ✨ Comment added

carolynvs:
@vdice validate examples isn't returning a response back to github. Do you think it's related to this PR?

vdice:
> @vdice validate examples isn't returning a response back to github. Do you think it's related to this PR?

Thank you; yea, something still not right with the runSuite logic in brigade.js... looking/fixing...

vdice:
@carolynvs fixed now.

PR #895, opened 2020-05-04
chore(examples): update py/pip to v3 to appease azure-cli build

vdice:
Updates the Dockerfile for the aks-terrraform and wordpress-mysql example bundles to use Python v3 utilities, as required by recent versions of azure-cli.

Fixes example build failure as seen in https://brigadecore.github.io/kashti/jobs/validate-examples-01e78qnvf6v093x69qmkw7h3sx


carolynvs:
The build is failing because the other example for wordpress-mysql also needs this fix.

vdice:
Oops, thank you @carolynvs .  Similar lines in the Dockerfile for the `wordpress-mysql` bundle have been updated now.

PR #900, opened 2021-02-05
ci(Makefile): pin ajv to known working version

vdice:
* Pins the ajv cli we use for schema validation to a known working version; fixes the validate-examples CI check


carolynvs (approved):
You are the best, thanks Vaughn!

PR #901, opened 2021-07-25
modified example cnab/app/run script in bundle-guide.md

vijaynag-bs:
added '$' for variable reference at line 24 and 26.


carolynvs (approved):
Heads up that duffle isn't actively maintained anymore. If you are looking for a replacement, I recommend https://porter.sh

vdice (approved):

